PR #6, opened 2015-01-17
FIX (typo) use dash in name

### LINKED ISSUE #3: Decide on the package name
Currently, @hayd changed the package name to `pandas_datareader`. Is everybody satisfied with that? (to be clear: it is about the name that is `import`ed)

Personally, I find it a bit long, and I think I also don't really like the underscore in the name. But of course, it is just personal taste! (I justed wanted to bring up the discussion to have this now, and not later, and deliberately decide on this to keep it as it is or to change it). 
What about just `datareader`, or `pddatareader` (but that is a bit difficult with the two 'd's)

hayd:
fixes #3 (my typo)


hayd:
Well, this doesn't work.

(Nosetest doesn't actually run any tests if there's a space in the package dir name ?)

PR #10, opened 2015-01-22
TST: Make tests work on python 2.6

### LINKED ISSUE #9: TST: make sure tests work on python 2.6

dstephens99:
Fixes #9


hayd:
Thanks for fixing that, my bad for leaving that in a bad state. Sorry!

hayd:
(I was surprised that the pandas tests no longer have `self.assertIsInstance` this used to be overridden in `pandas.util.testing` which was useful for 2.6 - ... The other option here is to use unittest2 in python 2.6 - I wonder if pandas should do this or should extract out pandas_testing)

*I had previously located the PR, where this was removed/made `tm.assertIsInstance` so it's no longer same API across pandas versions :(, can't find it atm.*

dstephens99:
No prob.  I was the one that wrote those incompatible tests in pandas..

PR #11, opened 2015-01-22
BUG:Fix options strike price float conversions if comma.

### LINKED ISSUE #7: Options strike price float conversions
pandas issue here: https://github.com/pydata/pandas/issues/9010, with linked PR

dstephens99:
Fixes #7 
Based on PR: pydata/pandas#9198

### DIFF
--- a/pandas_datareader/data.py
+++ b/pandas_datareader/data.py
@@ -698,8 +698,19 @@ class Options(object):
 
     def _get_underlying_price(self, url):
         root = self._parse_url(url)
-        underlying_price = float(root.xpath('.//*[@class="time_rtq_ticker Fz-30 Fw-b"]')[0]\
-            .getchildren()[0].text)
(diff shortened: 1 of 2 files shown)


pandas_datareader/data.py
```diff
@@ -698,8 +698,19 @@ def _option_frames_from_url(self, url):
...
+        underlying_price = root.xpath('.//*[@class="time_rtq_ticker Fz-30 Fw-b"]')[0]\
+            .getchildren()[0].text
+
+        try:
+            underlying_price = float(underlying_price)
+        except ValueError:
+            # check for comma
+            underlying_price = underlying_price.replace(',', '')
+
+        try:
```
hayd: Note: You could probably indent this block so float is only applied once is the usual case, or always replace (which may be cleaner).

dstephens99:
@hayd Yes, I prefer the always replace for cleanliness.  I don't think the performance penalty would be that bad, given that this doesn't get called that often (once per underlying).

PR #14, opened 2015-01-23
FIX correct license, package name and authors/email

### LINKED ISSUE #3: Decide on the package name

hayd:
Fixes #3 the name to pandas-datareader.

Corrects the license, url and contact info etc. for setup.py.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,87 @@
+=======
+License
+=======
+
+pandas is distributed under a 3-clause ("Simplified" or "New") BSD
+license. Parts of NumPy, SciPy, numpydoc, bottleneck, which all have
+BSD-compatible licenses, are included. Their licenses follow the pandas
+license.
+
+pandas license
+==============
+
+Copyright (c) 2011-2012, Lambda Foundry, Inc. and PyData Development Team
+All rights reserved.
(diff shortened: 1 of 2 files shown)


hayd:
cc @jorisvandenbossche  @jreback does this look right/better?

jorisvandenbossche:
@hayd You now included the pandas licence. I have to say, I don't know much about licenses, but shouldn't there also be a license that is mentioning pandas-datareader?

dstephens99:
This looks fine to me (but I don't know anything about licences).  Is this good to go? @jorisvandenbossche @hayd @jreback

hayd:
It could probably be modified in future to something like pandas-data-reader follows the pandas license, for now this is an obvious improvement.

PR #25, opened 2015-03-25
BUG: _retry_read_url retries connection even if successful

evanpw:
This bug also makes the "raise IOError" line unreachable. You'll get an UnboundLocalError on rs instead, if all retries fail.


dstephens99:
Thanks!.

PR #29, opened 2015-03-27
BUG: Add timezone to quote times in Options.

### LINKED ISSUE #28: Options quotes are missing timezone
I ran into this problem on the 'Quote_Time' field when pulling options quotes using `pandas.io.data.Options.get_all_data()`. While 'Quote_Time' is objectively 'US/Eastern' timezone, it actually shows up in the dataframe as naive timezone, which in my use case then got interpreted as UTC. I would suggest setting the timezone correctly when the data is pulled from Yahoo.

I'm pretty sure that the Expiry field in the index also uses a naive timezone (`= None`).

[Here](https://github.com/aisthesis/pynance/issues/3) is a related issue for further clarification.

davidastephens:
Fixes #28

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ __pycache__
 .coverage
 build
 dist
+.idea
--- a/README.rst
+++ b/README.rst
@@ -1,6 +1,6 @@
-===== 
+=================
 pandas-datareader
-===== 
+=================
 
 Up to date remote data access for pandas, works for multiple versions of pandas.
 
@@ -20,7 +20,7 @@ Install via pip
    $ pip install pandas-datareader
 
 Usage 
---------
+-----
 
 Rather than import from ``pandas.io`` you should import from ``pandas-datareader``:
 
--- a/pandas_datareader/data.py
+++ b/pandas_datareader/data.py
@@ -730,10 +730,10 @@ class Options(object):
         #Gets the time of the quote, note this is actually the time of the underlying price.
         try:
             quote_time_text = root.xpath('.//*[@class="time_rtq Fz-m"]')[0].getchildren()[1].getchildren()[0].text
-            ##TODO: Enable timezone matching when strptime can match EST with %Z
             quote_time_text = quote_time_text.split(' ')[0]
             quote_time = dt.datetime.strptime(quote_time_text, "%I:%M%p")
             quote_time = quote_time.replace(year=CUR_YEAR, month=CUR_MONTH, day=CUR_DAY)
+            quote_time = to_datetime(quote_time).tz_localize('US/Eastern') #localize to US/Eastern
         except ValueError:
             quote_time = np.nan
 
--- a/pandas_datareader/data.py
+++ b/pandas_datareader/data.py
@@ -998,14 +998,14 @@ class Options(object):
 
         expiry_dates = self.expiry_dates
         expiry = to_datetime(expiry)
-        if hasattr(expiry, 'date'):
-            expiry = expiry.date()
+        if expiry.tz is None:
+            expiry = expiry.tz_localize('US/Eastern')
 
         if expiry in expiry_dates:
             return expiry
         else:
             index = DatetimeIndex(expiry_dates).order()
-            return index[index.date >= expiry][0].date()
+            return index[index >= expiry][0]
 
     def get_forward_data(self, months, call=True, put=False, near=False,
                          above_below=2):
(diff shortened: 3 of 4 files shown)


jreback:
setting the timezone is s bit odd
and mskes working with these times much harder 
and are generally not as efficiently represented 

so unless you have a really compelling reason I would not put the tz

dstephens99:
Ya, I'm on the fence about the second one.  Was unaware that it wasn't efficiently represented.  I think we just have timezone on the quote time (first commit).

jreback:
look at the dtypes of the frame before/after

dstephens99:
I see - datetime64 vs object.  Looks like it doesn't matter if its in the index (both DateTimeIndex)?

I'm fine with leaving them naive.  Thoughts @aisthesis?

aisthesis:
The problem with naive is that applications will typically just interpret it as UTC, which it definitely isn't in the case of quote time. I feel like quote time at least should have a clear timezone. For performance in getting the data, the overhead in retrieval looks to me like it's coming from connecting to various Yahoo sources and not from Pandas putting everything together. But I could be wrong there, as I only see it as consumer.

It's also clear that the expirations are in a different spot both substantively (there's no actual time but just a date) and programmatically (they're a Pandas Timeseries within the index whereas quote time is a numpy datetime64 because numpy really only understands floats inside the array).

What do you guys think of making quote time have a timezone but leaving expiration tz-naive?

aisthesis:
I disagree with jreback that setting the times is odd and most definitely that it makes working with these times more difficult for the user. Speaking as user, it creates problems for me not having the timezone. If you decide to leave it naive, I'll have to put in my own code to set the timezone in both cases. I can't speak to the internal efficiency issue within Pandas.

dstephens99:
3 options:
1) Leave both naive
2) Add timezone to quote time
3) Add timezone to expiry date & quote time.

I'm inclined to the 2nd one.  The timezone for the quote time has a meaning and would be useful for people not on the east coast.  I'm iffy on adding a timezone to the expiry date.  However, if we start adding option quotes for European exchanges, then timezones for expiry dates might start mattering.

@jreback @aisthesis Thoughts?

jreback:
you have to be be consistent across the various methods. E.g. if you choose to do a timezone on say quote time, then the other data methods should do the same (I mean for say stock data). But to be honest I suspect most people either keep this data as relative and naive (e.g. 4pm, e.g. the reported time but as a naive time), or convert to UTC.

aisthesis:
@jreback I suspect very few do anything with it as naive. You run into problems very quickly, as I have, unless your local machine doing the processing happens to be on the U.S. East Coast. Naive isn't a problem as long as you're just doing something like pulling it from command line to check it out. But as soon as the date matters in a program, you're forced (as I've done) simply to insert the proper timezone. Moreover, for those who keep the data as naive timezone, it wouldn't matter if it were correctly specified. Whereas not specifying it definitively creates problems for some. In other words, even if it doesn't particularly matter for some users, specifying the timezone hurts no one and helps most users who actually use the time programmatically.

Yahoo! presumably is just providing it as a string with no tz specification, right?

aisthesis:
Maybe the least common denominator would be to put everything in UTC. The problem I have with naive timezones is that there is no reliable contract as to what the time actually is. So, I can't write reliable client code comparing times. Common use cases: Has the option expired? Is the given `Quote_Time` at market close? This will become a horrible mess if you ever expand the library beyond U.S. markets where then we would have naive times from clearly distinct timezones.

I can live with any of the following solutions:

1. Everything has a timezone and correct time for that zone. I'm ok with 'UTC' or 'US/Eastern' for U.S. markets. I agree with @jreback that the choice should be consistent.
2. Timezone is None but there is a clear contract that times are actually UTC. I find this suboptimal, but then I know what I'm dealing with and can write my code accordingly.

I find it unsatisfactory to have a naive timezone with no way to tell what the corresponding UTC time actually is.

davidastephens:
Just to continue this discussion.   What if we just make objects that have a time attached to them (ie: just option quote time currently) have the correct time zone, while leaving objects that are just dates naive.  Does that work? Or do we need to be all or none?

aisthesis:
That would work for my use case.

aisthesis:
What was the final verdict here? Are the Timestamps all in EST now?

davidastephens:
They are still all naive.  @jreback is your concern about performance alleviated now in 0.17+?

http://pandas.pydata.org/pandas-docs/stable/whatsnew.html#whatsnew-0170-tz

jreback:
yep putting them in tz would be ok now
(though an API change)

PR #31, opened 2015-03-27
DOC small formatting change in readme

jorisvandenbossche:
@dstephens99 Currently the readme is not nicely formatted on PyPI. And it is possible that it is just because of this formatting 'error'. 

Can you try this?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,6 +1,6 @@
-===== 
+=================
 pandas-datareader
-===== 
+=================
 
 Up to date remote data access for pandas, works for multiple versions of pandas.
 
@@ -20,7 +20,7 @@ Install via pip
    $ pip install pandas-datareader
 
 Usage 
---------
+-----
 
 Rather than import from ``pandas.io`` you should import from ``pandas-datareader``:
 


dstephens99:
Sure.  I'm not that familiar with Pypi - but it seems like I can't change a release, so we'd have to do a v0.1.1 release in order to update this.  Unless there's a way to change it that I don't know about?

jorisvandenbossche:
I think that you can just run `python setup.py register` again to update the metadata (but not sure, never done it myself)

hayd:
You can just edit the manually on pypi (when logged in), no reuploading necessary. It *may* be because pypi doesn't support svg images, changing the svg to png in the badges may fix it.

dstephens99:
Thanks - I edited the RST.  Works now (SVG is fine).

hayd:
>  (SVG is fine).

pypi have fixed a bug, awesome!

bashtage:
There appears to be a non-printing character in example

https://pypi.python.org/pypi/pandas-datareader/0.1.0

dstephens99:
@bashtage Thanks.  Fixed in #32

Issue #33, opened 2015-03-28
TST: Test class TestGoogle never runs due to bug in pandas.util.get_locales


davidastephens:
pydata/pandas#9744

hayd:
if we get coverage to 100% it should help prevent this in future, you can see what's being skipped: https://coveralls.io/builds/2213917

dstephens99:
That's how I noticed it - I was looking at the coverage to try to improve it.   When I get this fixed, it should improve a lot.

davidastephens:
Fixed with #35

PR #35, opened 2015-03-28
COMPAT: Fix dl_mult_symbols for python3

dstephens99:


dstephens99:
This doesn't get tested currently due to #33, however, will need this fix when the bug is fixed in Pandas that allows TestGoogle to run.

PR #43, opened 2015-04-19
add quandl as a source for historical financial data

briancappello:
Replaces https://github.com/pydata/pandas/pull/9935

| Q |	A |
------------ | -----
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Tests pass? | yes
| Fixed tickets | -
| License | BSD
| Doc PR | -

### DIFF
--- a/pandas_datareader/tests/test_data.py
+++ b/pandas_datareader/tests/test_data.py
@@ -330,7 +330,7 @@ class TestYahooOptions(tm.TestCase):
         self.assertTrue(len(options) > 1)
 
     def test_options_is_not_none(self):
-        option = web.Options('aapl')
+        option = web.Options('aapl', 'yahoo')
         self.assertTrue(option is not None)
 
     def test_get_call_data(self):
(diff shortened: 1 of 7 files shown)


pandas_datareader/tests/test_data.py
```diff
@@ -135,6 +135,95 @@ def test_unicode_date(self):
         self.assertEquals(data.index.name, 'Date')
 
 
+
+class TestQuandl(tm.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        print('hiiiiii')
```
dstephens99: Should drop this line..

dstephens99:
This looks great to me.  Can you add to the docs?

davidastephens:
@briancappello Just following up on this pull-request.  Can you drop the print('hiiii') line in the tests and add this to the docs, then we can merge?

femtotrader:
What about http://pythonhosted.org/Quandl 

moreover it seems that API v1 will be deprecated

https://www.quandl.com/help/api

(Note: The guide to v1 of the Quandl API is available here. We encourage users to adopt the current version, v1 is on the path to deprecation.)

briancappello:
@davidastephens Apologies on dropping the ball here. I just pushed up some commits to address the outstanding issues raised; the debug print statement has been removed, I updated to using the quandl api v3, and added basic documentation.

davidastephens:
Thanks @briancappello.  Can you rebase and squash?

@bashtage @femtotrader   Are you guys good with this?  Seems like a reasonable way to provide a consistent API between yahoo, google and quandl.

femtotrader:
We shouldn't merge

see https://github.com/briancappello/pandas-datareader/tree/quandl_historical_data_source/pandas_datareader

all is in `data.py`

femtotrader:
I also wonder differences with https://github.com/quandl/quandl-python/tree/feature-AP-243/v3-client

briancappello:
@davidastephens @femtotrader I refactored this to match the new architecture. Tests are failing because quandl limits non-authenticated users to 50 API calls per day (there are less than 50 calls but I had pushed maybe an hour ago causing this run to fail). Does nose have something akin to "try to run this test but skip it if it fails" or maybe there's a better way to handle this class of error?

@femtotrader I haven't played with the official quandl package much; at first glance it seems to lack support for downloading more than one symbol at once, which we support here.

bashtage:
You can tell nose that it is ok if a test fail.An example of a decorator that implements this is here: 

http://stackoverflow.com/questions/9613932/nose-plugin-for-expected-failures

50 API calls/day is a problem for testing IMO.  You should probably get an API key from quandl and follow 

http://docs.travis-ci.com/user/encryption-keys/

to insert an encrypted version into the main.  This will at least allow the test to always run on travis.

bashtage:
As a comment I don't like the setting your API key as an environmental variable.  Why not let the api key be set at the module level and kept in a module level variable?  This would require a dedicated function like `set_quandl_api_key('xxxx')`

briancappello:
@bashtage Thanks for the feedback! I added a set_api_key function to do (I think?) what you're talking about. As for the 50 calls/day being a problem, I agree with you there. Reading through the travis doc you linked, it mentions encrypted environment variables are not available for external pull requests (like, for example, this one or any other coming from a fork). I'm not very familiar with travis but it doesn't seem to support an approach other than environment variables for encrypted data?

femtotrader:
@bashtage, @briancappello is right.

With Travis, we need to be able to set API key using an environnement variable.

But code should be able to support both
- environnement variable
- passing an API key parameter

We should probably have a parameter called `credentials`.
This parameter could be a dict and `api_key` could be one of its keys.

Quandl only need an API key but some other datareader could need `login` / `password`...

so this `credentials` parameter could also allow this:

    start = datetime(2010, 1, 1)
    end = datetime(2013, 1, 27)
    df = web.DataReader('F', 'quandl', start, end, credentials=dict(api_key='....'))

or if we want a more oriented object approach

    df = web.DataReader('F', 'quandl', start, end, credentials=Credential(api_key='....'))

but if `credendials` is `None` or if no `'api_key'` is found, environnement variable should be use.

If no environnement variable for Quandl is found, query will be performed without setting API key.

femtotrader:
An other approach could be to have

    web.set_credentials(api_key='....')

so it could also allow (for other paid datareader)

    web.set_credendials(login='...', password='...')

but anyway I won't expose like you did a function like

    web.set_quandl_api_key('your_key_here')

because I think we should avoid to expose specific datareader function at this level.

davidastephens:
Did we ever get anywhere on adding the API key? @femtotrader @briancappello @bashtage

femtotrader:
I think we should close this PR. There is a Quandl Python library which is able to output Pandas DataFrame. see https://github.com/quandl/quandl-python and https://github.com/quandl/quandl-python/tree/feature-AP-243/v3-client2

PR #45, opened 2015-05-09
Extended DataReader to fetch Dividends and Stock Splits from Yahoo!

brotchie:
- Added yahoo-actions data_source to DataReader;
 - Added get_data_yahoo_actions that returns a DataFrame containing
   DIVIDEND and SPLIT corporate actions fetched from Yahoo! Finance;
 - Added unit tests;
 - Added documentation example.

### DIFF
--- a/docs/source/remote_data.rst
+++ b/docs/source/remote_data.rst
@@ -36,6 +36,8 @@ It should be noted, that various sources support different kinds of data, so not
 Yahoo! Finance
 ==============
 
+Historical stock prices from Yahoo! Finance.
+
 .. ipython:: python
 
     import pandas_datareader.data as web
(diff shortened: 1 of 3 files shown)


dstephens99:
Thanks.  Could you add that to the docs? I assume that would be a common question.

Only other thing I could think of would be to have a convenient way to get only the splits or only the dividends (understanding that anyone could filter the resulting dataframe).  Could be next version..

davidastephens:
@brotchie  Can you add the note to the docs about the date being the ex-date?  Then I'm good to merge.

Thanks!

brotchie:
@davidastephens I've updated the appropriate docstrings and Sphinx documentation with details of ex-dates.

Cheers!

davidastephens:
Great.  Thanks!  Can you squash and I'll merge when green?

PR #52, opened 2015-07-15
Update travis

bashtage:
Update Travis to use faster docker build.  Also include pandas 0.16 in testing.


bashtage:
Need to rebase to remove extra commit.

Issue #58, opened 2015-08-21
One file per datareader


femtotrader:
Hello,

`data.py` is a very (too) big file.

I think we should have one file per datareader.

Moreover this will help https://github.com/pydata/pandas-datareader/issues/48

I've done on my fork a branch where each datareader is a file.

    $ nosetests -s -v

shows

    Ran 58 tests in 83.806s
    FAILED (SKIP=7, errors=10, failures=3)

that's exactly same result with this `master` branch.

I will send a PR.

Kind regards

bashtage:
The splitting seems too fine grained IMO.  Why not split by data vendor only?

bashtage:
I think the utility functions, which is anything that isn't vendor specific, should also be collected in shared, rather than having one file per function.

femtotrader:
what utility function will you put in "shared" ?
I will personally move function to "shared" only if they are actually shared. 

I used this fine grained split because it will be easier for me to add caching (see https://github.com/pydata/pandas-datareader/issues/48 ) this way like I did in https://github.com/femtotrader/pandas_datareaders_unofficial .

Moreover it will be easier to see what is done and what need to be done.

femtotrader:
A vendor can decide to deprecate one API an not other ones.
Is Yahoo Finance Components deprecated ?
With one file per reader it will be IMHO easier and could also avoid spaghettis.

bashtage:
I've never some across a package that has one function per file (or anything close).  With modern editors it doesn't matter how large the files is since you can always use folding to hide anything you don't want to see. 

In terms of caching, is there some reason now to use django-style `@cache` decorator that assumes the exact same set of inputs produces the same output, so that only the final result is cached, and the reader doesn't need to be modified?

davidastephens:
I'm fine with splitting by API.  I agree with @bashtage though, I don't think we need one file per shared function.   Just one shared file works.

Any benefit of using subpackages by vendor?  
Ie: pandas_datareader/yahoo/options.py 
Ie: pandas_datareader/yahoo/actions.py

bashtage:
@davidastephens subpackage suggestion looks reasonable and seems to be more canonical Python packaging.

So something like pdr/vendor/datatype plus pdr/common.py where common is anything that isn't vendor specific.

PR #59, opened 2015-08-21
One file per datareader

femtotrader:
Should fix https://github.com/pydata/pandas-datareader/issues/58


femtotrader:
@davidastephens @bashtage Could you have a look ?

pandas_datareader/data.py
```diff
@@ -84,1225 +69,11 @@ def DataReader(name, data_source=None, start=None, end=None,
                                       retry_count=retry_count, pause=pause)
     elif data_source == "google":
         return get_data_google(symbols=name, start=start, end=end,
-                               adjust_price=False, chunksize=25,
-                               retry_count=retry_count, pause=pause)
+                               chunksize=25, retry_count=retry_count, pause=pause)
```
davidastephens: Why is adjust_price removed?

davidastephens:
Do we need the datareaders package? Can it just be pandas_datareader/google/..?

femtotrader:
I created the `datareaders` package because I didn't like the fact of having for example `fred.py` `famafrench.py`, `google` and `yahoo` in the same directory than `date_chunks.py`, `shared.py` and `url_request.py`

I can remove it if you think that's more understandable and put `date_chunks.py`, `shared.py` and `url_request.py` in a `commons` package.

`adjust_price` is removed because Google don't provide it (only Yahoo)
That's the reason why 

    if ret_index:
        hist_data['Ret_Index'] = _calc_return_index(hist_data['Adj Close'])
    if adjust_price:
        hist_data = _adjust_prices(hist_data)

is in `get_data_yahoo`

femtotrader:
@davidastephens I have update accordingly.

femtotrader:
I also think that some cleanup is / will be necessary.

For example, 

- in `famafrench.py`
  - `_FAMAFRENCH_URL` should be named `_URL`
  - `get_data_famafrench` should be name only `get_data`

most files should have such cleanup but maybe we should do it after merging.

`options.py` should also be improved because it's in `yahoo` so there is no need for `data_source``

we should have an other `options.py` at same level than `data.py`

femtotrader:
Options is now also a function in `data.py` which is able to dispatch per data vendor. For now there is only one data "vendor" for options : Yahoo.

davidastephens:
I think for now, given there is only 2 things in the common package, we can just have it as common.py in the base package.

davidastephens:
I think we should merge the package structure changes and do the clean-up and option dispach in 2 separate pulls.  

Can you rebase against master and squash to one commit (except the option dispatch and clean-up?).

Thanks for all the work on this.

femtotrader:
Hum... that will be quite difficult for me. I'm not very good with git. Maybe I should create an other branch ?

davidastephens:
There is a good tutorial here:
https://github.com/pydata/pandas/wiki/Using-Git

Here are the steps you need to do:
```
#add the pandas remote repository to your fork and name it upstream

git remote add upstream git://github.com/pydata/pandas-datareader

#fetch and then rebase interactively to squash, reword, fix, otherwise change some commits

git fetch upstream
git rebase --interactive upstream/master

```

In the text file you get from rebasing, just remove the Option and cleanup commits (delete the lines) and change the rest of them from pick to squash (except the first one).  That squashes them all into the first commit.

femtotrader:
Closing this PR is favor of https://github.com/pydata/pandas-datareader/pull/68

femtotrader:
Sorry

PR #65, opened 2015-08-22
CLN: Unnecessary try: block removed from Options

davidastephens:


davidastephens:
The property that this is calling has the same try: except block.

femtotrader:
I think we should have apply this PR after splitting code because we will have to apply it again.

PR #68, opened 2015-08-23
New package structure

femtotrader:
Should fix https://github.com/pydata/pandas-datareader/issues/58
PR https://github.com/pydata/pandas-datareader/pull/59/ should be ignored

PR #69, opened 2015-08-23
CLN: Split into subpackages

### LINKED ISSUE #58: One file per datareader

davidastephens:
@femtotrader @bashtage 

Everyone fine with this version?

Fixes #58, replaces #68.


hayd:
Personally I would prefer util or utils to common. But this is a big improvement overall. :+1:

davidastephens:
@hayd  Agreed, this is a big improvement - thanks @femtotrader.

I'll change to util..

Other than compatibility, data.py is redundant now.  May be worth changing it to api.py and importing everything to the top level.

hayd:
The data.py is for backwards compat. The old API should remain available for a good while. Though there are other ways to get the same behaviour. IMO without a deprecation message.

bashtage:
I agree that importing `data` should remain supported for the next few releases.

femtotrader:
I noticed that `google.daily` provide only daily data but `yahoo.daily` can provide other interval

maybe `yahoo.daily` should be named differently.

But thanks for merging.

Issue #73, opened 2015-08-24
Options dispatch in data.py doesn't work correctly when None is given as data_source


femtotrader:
Hi,

see

    def Options(symbol, data_source=None):
        if data_source is None:
            warnings.warn("Options(symbol) is deprecated, use Options(symbol,"
                            " data_source) instead", FutureWarning)
            data_source = "yahoo"
        elif data_source == "yahoo":
            return YahooOptions(symbol)
        else:
            raise NotImplementedError("currently only yahoo supported")

if `data_source` is `None` nothing will be returned because it's like a nested if construct

It should be 

        if data_source is None:
            warnings.warn("Options(symbol) is deprecated, use Options(symbol,"
                            " data_source) instead", FutureWarning)
            data_source = "yahoo"

        if data_source == "yahoo":
            return YahooOptions(symbol)
        else:
            raise NotImplementedError("currently only yahoo supported")

PR #74, opened 2015-08-24
Options dispatch in data.py shouldn't return None when data_source=No…

femtotrader:
…ne

should fix https://github.com/pydata/pandas-datareader/issues/73


davidastephens:
Thanks!

PR #84, opened 2015-08-29
ENH: Don't pause on first attempt.

### LINKED ISSUE #77: Pause in _retry_read_url should only happens from the second time
Hi,

I noticed in `utils.py` in `_retry_read_url` this

    def _retry_read_url(url, retry_count, pause, name):
        """
        Open url (and retry)
        """
        for _ in range(retry_count):
            time.sleep(pause)
            ...

So an unnecessary pause occured from begining (first request)

we might have something like:

    def _retry_read_url(url, retry_count, pause, name):
        """
        Open url (and retry)
        """
        for i in range(retry_count):
            if i != 0:
                time.sleep(pause)
            ...

davidastephens:
Fixes #77

### DIFF
--- a/pandas_datareader/utils.py
+++ b/pandas_datareader/utils.py
@@ -104,7 +104,6 @@ def _retry_read_url(url, retry_count, pause, name):
     Open url (and retry)
     """
     for _ in range(retry_count):
-        time.sleep(pause)
 
         # kludge to close the socket ASAP
         try:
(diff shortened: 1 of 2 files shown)


femtotrader:
I don't think that's a great idea to do a sleep at each end of loop because it will add an unnecessary pause at last attempt.

davidastephens:
It only does the sleep if it doesn't return.

femtotrader:
Oh yes sorry.

Issue #97, opened 2015-09-29
ImportError: No module named 'pandas_datareader.google'


femtotrader:
Hello,

I'm facing a a problem when I try a "clean" install of pandas-datareader.

    from pandas_datareader import data

raises

    ImportError: No module named 'pandas_datareader.google'

When I uninstall `pandas-datareader` I can notice that there is no "google" directory (neither "yahoo", neither "tests" but that's less problematic).

    pc:~ femto$ pip install git+https://github.com/pydata/pandas-datareader.git
    Collecting git+https://github.com/pydata/pandas-datareader.git
      Cloning https://github.com/pydata/pandas-datareader.git to /var/folders/j_/v8b1bst93_94t724ptsswfsr0000gn/T/pip-bzcvj335-build
    Requirement already satisfied (use --upgrade to upgrade): pandas in /anaconda/lib/python3.4/site-packages (from pandas-datareader==0.1.1)
    Requirement already satisfied (use --upgrade to upgrade): python-dateutil>=2 in /anaconda/lib/python3.4/site-packages (from pandas->pandas-datareader==0.1.1)
    Requirement already satisfied (use --upgrade to upgrade): pytz>=2011k in /anaconda/lib/python3.4/site-packages (from pandas->pandas-datareader==0.1.1)
    Requirement already satisfied (use --upgrade to upgrade): numpy>=1.7.0 in /anaconda/lib/python3.4/site-packages (from pandas->pandas-datareader==0.1.1)
    Requirement already satisfied (use --upgrade to upgrade): six>=1.5 in /anaconda/lib/python3.4/site-packages (from python-dateutil>=2->pandas->pandas-datareader==0.1.1)
    Installing collected packages: pandas-datareader
      Running setup.py install for pandas-datareader
    Successfully installed pandas-datareader-0.1.1
    pc:~ femto$ ipython
    Python 3.4.3 |Anaconda 2.3.0 (x86_64)| (default, Mar  6 2015, 12:07:41)
    Type "copyright", "credits" or "license" for more information.

    IPython 4.0.0 -- An enhanced Interactive Python.
    ?         -> Introduction and overview of IPython's features.
    %quickref -> Quick reference.
    help      -> Python's own help system.
    object?   -> Details about 'object', use 'object??' for extra details.

    In [1]: from pandas_datareader import data
    ---------------------------------------------------------------------------
    ImportError                               Traceback (most recent call last)
    <ipython-input-1-75f869254055> in <module>()
[...]

PR #98, opened 2015-09-29
Add google and yahoo module

femtotrader:
Should fix https://github.com/pydata/pandas-datareader/issues/97


davidastephens:
Thanks.

Issue #99, opened 2015-10-01
Use find_packages() to automatically find packages


femtotrader:
Hello,

in a recent issue I noticed some problem with package
https://github.com/pydata/pandas-datareader/issues/97

I submited a PR https://github.com/pydata/pandas-datareader/pull/98
which have been merged by @davidastephens 

I was replacing

    packages=['pandas_datareader'],

by

    packages=['pandas_datareader',
        'pandas_datareader.google', 'pandas_datareader.yahoo'
    ],

I wonder why not doing

     from setuptools import setup, find_packages

and

    packages=find_packages(exclude=['contrib', 'docs', 'tests*']),

So if we add other package we won't have to modify `setup.py`

Your opinion ?

Kind regards

bashtage:
This is the preferred way to do things once a package becomes even moderately large.

PR #100, opened 2015-10-02
FIX find packages automatically

femtotrader:
Try to fix https://github.com/pydata/pandas-datareader/issues/99


davidastephens:
Thanks!

PR #110, opened 2015-10-17
CLN: Cleanup subclass

### LINKED ISSUE #106: Cleanup subclass structures
There are few options commonly used in some readers, such as proxy and cache(#48). How about:

- Change read functions to classes which has common base class.
- Use ``requests`` to get raw files rather than ``urlopen``.

sinhrks:
Closes #106. Clean-up most of classes to use common logic.


femtotrader:
Your `BaseReader` should probably have a `_get_session` method

```
def _get_session(self, session):
    """Returns a requests.Session if session is None
    or session if it's not None (cached session
    with requests-cache for example)
    :param session:
    :return:
    """
    if session is None:
        session = requests.Session()
    return session
```

and `__init__` should accept a `session` parameter (default to `None`)

so we will do

    self.session = _get_session(session)

this will allow to pass a session (with proxy or cache)

sinhrks:
OK, updated to add a ``session``, I've chosen a word "init" as it is not for getting.

BTW, found a problem that ``requests`` doesn't allow to specify pausing between retry. Thus, current ``pause`` kw can't work now... I'll implement retry-pause logic by myself.

sinhrks:
Implemented the logic to handle retry and pause. Now ready for review.

femtotrader:
Sorry but I don't see commit

https://github.com/sinhrks/pandas-datareader/tree/cln_subclass

sinhrks:
It's squashed to the single commit, following pandas manners.

femtotrader:
Ok yes sorry... I was mistaken by commit date.

bashtage:
Is anything in `base` for public use?  If not private seems more correct.

sinhrks:
@bashtage No, changed it to private. Rebased and squashed.

davidastephens:
Thanks @sinhrks  This looks great.

Issue #119, opened 2015-11-16
Code cleanup (unused imports, unused variables...)


femtotrader:
Hello,

according to 
https://landscape.io/github/pydata/pandas-datareader/
there is 2 "errors" (same line 75 in wb.py)

    def download(country=['MX', 'CA', 'US'], indicator=['NY.GDP.MKTP.CD', 'NY.GNS.ICTR.ZS'], ...)

these 2 errors are

    Dangerous default value [] as argument

There is also some "smells" and some "Style" issues.

Badge [![Code Health](https://landscape.io/github/pydata/pandas-datareader/master/landscape.svg?style=flat)](https://landscape.io/github/pydata/pandas-datareader/master)

can be inserted with the following reStructuredText code

    .. image:: https://landscape.io/github/pydata/pandas-datareader/master/landscape.svg?style=flat
       :target: https://landscape.io/github/pydata/pandas-datareader/master
       :alt: Code Health

Kind regards

femtotrader:
Dangerous default value [] error 

is fixed by http://github.com/pydata/pandas-datareader/pull/121

but some cleanup is still possible

femtotrader:
It may be a good idea to add `flake8` to `.travis.yml`

femtotrader:
This PR https://github.com/pydata/pandas-datareader/pull/162 being merged we can probably close this issue now.

PR #121, opened 2015-11-22
CLN: Cleanup WorldBank reader

### LINKED ISSUE #120: wb.py needs a rewrite
Hello,

I think that all DataReaders are now objects except `wb.py` maybe this file need a rewrite.

see also https://github.com/pydata/pandas-datareader/issues/117

It should help to have cache support for World Bank https://github.com/pydata/pandas-datareader/issues/48

One possible API usage could be

    import requests_cache
    import datetime
    session = requests_cache.CachedSession(cache_name='cache', backend='sqlite', expire_after=datetime.timedelta(days=60))
    DataReaderObject("wb", session=session).search(string='gdp.*capi', field='name', case=False)
    DataReaderObject("wb", session=session).get_indicators()

Kind regards

sinhrks:
Closes #120 to remove ``urlopen``. 

Related #119, changed some default arg using ``list`` to ``None`` (and set them in the method).

 CC @femtotrader

### DIFF
--- a/pandas_datareader/_utils.py
+++ b/pandas_datareader/_utils.py
@@ -1,5 +1,4 @@
-import warnings
-
+import pandas as pd
 from pandas.core.common import PandasError
 
 
@@ -10,3 +9,21 @@ class RemoteDataError(PandasError, IOError):
     pass
 
 
+from distutils.version import LooseVersion
+
+PANDAS_VERSION = LooseVersion(pd.__version__)
+
+if PANDAS_VERSION >= LooseVersion('0.17.0'):
+    PANDAS_0170 = True
+else:
+    PANDAS_0170 = False
+
+if PANDAS_VERSION >= LooseVersion('0.16.0'):
+    PANDAS_0160 = True
+else:
+    PANDAS_0160 = False
+
+if PANDAS_VERSION >= LooseVersion('0.14.0'):
+    PANDAS_0140 = True
+else:
+    PANDAS_0140 = False
--- a/pandas_datareader/base.py
+++ b/pandas_datareader/base.py
@@ -7,7 +7,7 @@ import requests
 
 from pandas import to_datetime
 import pandas.compat as compat
-from pandas.core.common import PandasError
+from pandas.core.common import PandasError, is_number
 from pandas import Panel, DataFrame
 from pandas import read_csv
 from pandas.compat import StringIO, bytes_to_str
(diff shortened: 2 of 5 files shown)


femtotrader:
You might pass session parameter to each function which should download data in order to also provide cache mechanism to WorldBank data reader. (CC @davidastephens )

An other rewrite (maybe in a next release) will be to create a class.

femtotrader:
Sorry. In fact, there is now a class `WorldBankReader` which is a great improvement.

but you might to also pass session to `download`, `get_countries`, `get_indicators`, `search` and send this session parameter to `WorldBankReader`

femtotrader:
An other improvement might be about `errors` parameter.

We should use string but Enum (enum34 for Python 2) as it wil allow tab completion.

femtotrader:
A last idea (about methods of WorldBankReader)...

`get_countries` and `get_indicators` might be properties named `countries` and `indicators` are there is no parameter (except `self`) for these 2 WorldBankReader methods.

femtotrader:
I thought that it was the last idea... but sorry I also noticed that there is a `per_page` parameter in urls.

We might support chunks and use `page` parameter

try
http://api.worldbank.org/countries/?per_page=1000000&format=json&page=1

and

http://api.worldbank.org/countries/?per_page=10&format=json&page=1
http://api.worldbank.org/countries/?per_page=10&format=json&page=2

but this is probably out of the scope of this PR and an issue should be opened for this.

sinhrks:
@femtotrader Thanks for the comments. I've updated the PR as below. I intend this PR to cover existing functionality, and others should be handled separately.

- **Add ``session`` to WB's ``get_xxx`` funcs:** Done. As other kw such as retry may be useful, passed all kwds to ``WorldBankReader`` class (same as other ``get_data_xxx`` funcs)

- **Change ``get_indicators`` and ``get_countries`` to property:** Don't do. I don't think the names are very understandable for users which is different from plain ``get_xxx`` functions. 

- **Change ``errors`` to enum:** Don't do in this PR. Because ``pandas`` doesn't use enum. Pls prepare a PR separately.

- **Add ``page`` param:** Don't do in this PR.

femtotrader:
Thanks a lot for you work. You are right! Other improvements should be done in an other PR.

femtotrader:
Sorry @sinhrks I think that ther is a mistake with

    return WorldBankReader().get_countries(**kwargs)

it might be

    return WorldBankReader(**kwargs).get_countries()

sinhrks:
@femtotrader Thanks! Fixed and added tests.

PR #122, opened 2015-11-23
ENH: Add requests session to Options

davidastephens:
Adds requests session to Options, increases options parsing by 4x by skipping read_html.


femtotrader:
I didn't know `requests-file`

https://pypi.python.org/pypi/requests-file
https://github.com/dashea/requests-file

davidastephens:
Yes, I needed to use it in order to use the requests functions on the test examples.

Issue #48, opened 2015-06-26
Support caching


twiecki:
Are there any plans to add caching of downloaded files?

femtotrader:
In https://github.com/femtotrader/pandas_datareaders_unofficial I was using requests-cache http://requests-cache.readthedocs.org/

but @bashtage seems to also have some ideas about it. He's suggesting the use of Django cache decorator

maybe we might weight here the pros and cons of each method.

bashtage:
I'm not literally suggesting using django's cache - I was suggesting using their strategy, which is really algorithmically simple - same function, same inputs, read cache, else read fresh (in django it is a sql read, here it will be a network read).  This would remove an install dependency on a 3rd party caching solution.

femtotrader:
Sorry @bashtage (english is not my mother tongue)

My idea was to pass a `session` parameter (with `None` as default value). If a requests-cache session is passed it will use it. If no session is passed a normal `requests` is used. But yes it adds a dependency to a library (`requests` for "normal" session and an optional dependency to `requests-cache` if we want to cache results).

What I don't like with requests-cache (for now) is the lack of support of several database https://github.com/reclosedev/requests-cache/issues/43

femtotrader:
I'm working on this here:

https://github.com/femtotrader/pandas-datareader/tree/session

Cache seems to work fine with 
- Google Daily
- Yahoo Daily

I think that all other datareaders have a `session` parameter (set to `None` by default)
we just need to call

    session = _get_session(session)

to get a requests.Session

and use it inside other datareaders

Some help will be nice.

twiecki:
That looks great. Will it automatically try to update the cache if I request data that's not available?

davidastephens:
@twiecki  I think how it works is if you make a request to a website, it will check if its in the cache and not expired, if not it will get a fresh copy from the server.

davidastephens:
@femtotrader  I'm not sure we need to test to make sure the cache is working, requests-cache isn't a dependency.  We should probably test that the session parameter is working though - any way to confirm with a test that you used the same requests session?

I put together a list of the things we still need to do for this issue, please add to it if I missed anything.
- [ ] Tests to confirm session is working for each data reader
- [ ] Add session to famafrench.get_available_datasets
- [ ] Others?

PR #126, opened 2015-11-23
DOC: use of session parameter to cache queries

femtotrader:
Add a `cache.rst` file to explain how to cache queries.

Might help to fix https://github.com/pydata/pandas-datareader/issues/48


davidastephens:
Thanks!  I made a couple edits in #129

PR #136, opened 2015-12-06
Add session parameter to get_available_datasets

femtotrader:
Should help https://github.com/pydata/pandas-datareader/issues/48

### DIFF
--- a/pandas_datareader/_utils.py
+++ b/pandas_datareader/_utils.py
@@ -27,8 +27,3 @@ if PANDAS_VERSION >= LooseVersion('0.14.0'):
     PANDAS_0140 = True
 else:
     PANDAS_0140 = False
-
-def _init_session(session):
-    if session is None:
-        session = requests.Session()
-    return session
--- a/pandas_datareader/_utils.py
+++ b/pandas_datareader/_utils.py
@@ -27,3 +27,8 @@ if PANDAS_VERSION >= LooseVersion('0.14.0'):
     PANDAS_0140 = True
 else:
     PANDAS_0140 = False
+
+def _init_session(session):
+    if session is None:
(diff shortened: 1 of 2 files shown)


pandas_datareader/famafrench.py
```diff
@@ -6,27 +6,29 @@
...
+    session = _init_session(session)
     try:
-        from lxml.html import parse
+        from lxml.html import document_fromstring
     except ImportError:
         raise ImportError("Please install lxml if you want to use the "
                           "get_datasets_famafrench function")
 
-    root = parse(_URL + 'data_library.html')
+    response = requests.get(_URL + 'data_library.html')
```
davidastephens: Don't you have to do session.get(_URL...) in order to use the session?

femtotrader: Yes you are right... sorry!

pandas_datareader/_utils.py
```diff
@@ -26,4 +26,9 @@ class RemoteDataError(PandasError, IOError):
 if PANDAS_VERSION >= LooseVersion('0.14.0'):
     PANDAS_0140 = True
 else:
-    PANDAS_0140 = False
\ No newline at end of file
+    PANDAS_0140 = False
+
+def _init_session(session):
```
sinhrks: Should use the logic as current ``_BaseReader``, and ``_BaseReader`` can use this definition.

https://github.com/pydata/pandas-datareader/blob/master/pandas_datareader/base.py#L57

femtotrader:
Yes I should do something like you did in WordBankReader

https://github.com/pydata/pandas-datareader/blob/master/pandas_datareader/wb.py

PR #145, opened 2015-12-12
Adds retrieval of EDGAR master index.

### LINKED ISSUE #143: ENH: Support retrieval of SEC EDGAR indices.
The SEC has a large database of company filings (e.g., annual reports, proxy statements), called [EDGAR](http://www.sec.gov/edgar/searchedgar/companysearch.html), that it makes publicly accessible. They also provide a set of (huge) [indices](http://www.sec.gov/edgar/indices/fullindex.htm) to all of those filings, and it would be really helpful to be able to pull that index, filter it down to a set of companies, date ranges, and filing types, and then use the filename fields to pull the data.

As a first step, I'd like to propose support for pulling an index file from the SEC, parsing it, and returning it as a DataFrame. I plan on working on it myself, so I'm mostly interested in whether it's a good fit. I think it so, but I'd like to gauge interest and solicit some feedback.

The data is different than a lot of stuff in `pandas_datareader` in that it's a step toward pulling text in addition to numeric data. That said, a lot of the research or data projects (mine for sure) that use the financial and industry data available with `pandas_datareader` may also be interested in this kind of data (or descriptive stats of the filings in the indices).

Later on, I think it would be nice to support pulling documents themselves, but I'd like to start with the indices for now.

jtkiley:
Closes #143.

Replaces #144.

### DIFF
--- a/pandas_datareader/edgar.py
+++ b/pandas_datareader/edgar.py
@@ -1,0 +1,43 @@
+from pandas import read_csv
+from pandas.io.common import ZipFile
+from pandas.compat import StringIO
+from pandas.compat import BytesIO
+
+from pandas_datareader.base import _BaseReader
+
+
+_URL_FULL = 'ftp://ftp.sec.gov/edgar/full-index/master.zip'
+_COLUMNS = ['cik', 'company_name', 'form_type', 'date_filed', 'filename']
+
+
+class EdgarIndexReader(_BaseReader):
+    """
+    Get master index from the SEC's EDGAR database.
+
+    Returns
+    -------
+    edgar_index : pandas.DataFrame.
+        DataFrame of EDGAR master index.
+    """
+
+    @property
+    def url(self):
+        return _URL_FULL
+
+    def _read_zipfile(self, url):
+
+        zipf = BytesIO(self._get_response(url).content)
+
+        with ZipFile(zipf, 'r') as zf:
+            data = zf.open(zf.namelist()[0]).read().decode()
+
+        return data
+
+    def _read_one_data(self, url, params):
+
+        index_file = StringIO(self._read_zipfile(url))
+
+        index = read_csv(index_file, delimiter='|', header=None,
+                         index_col=False, skiprows=10, names=_COLUMNS,
+                         low_memory=False)
+        return index
--- a/docs/source/remote_data.rst
+++ b/docs/source/remote_data.rst
@@ -396,3 +396,16 @@ You can specify dataset ID "tran_sf_railac" to get corresponding data via ``Data
 
     df = web.DataReader("tran_sf_railac", 'eurostat')
     df
+
+.. _remote_data.edgar:
+
+EDGAR Index
+===========
+
+Company filing index from EDGAR (SEC).
+
+.. ipython:: python
+
+    import pandas_datareader.data as web
+    ed = web.DataReader('full', 'edgar-index')
+    ed[:5]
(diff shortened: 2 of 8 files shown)


jtkiley:
I think I covered all of the bases (it works, tests, docs, whatsnew), but let me know if you'd like me to fix anything. Otherwise, let me know and I'll squash it.

pandas_datareader/edgar.py
```diff
@@ -0,0 +1,72 @@
...
+        Parameters
+        ----------
+        url : str
+            target URL
+        params : dict or None
+            parameters ignored, as it's FTP.
+        """
+
+        for i in range(self.retry_count + 1):
+            response = urlopen(url).read()
```
femtotrader: You should use 

    self.session

because it will become possible to have a cache mechanism to avoid several requests to same server

see https://github.com/pydata/pandas-datareader/issues/48

jtkiley: It looks like `requests` only supports HTTP, and the index is only available by FTP. I agree that it would be great to cache it, as I think it's only updated once per business day.

My initial look suggests that at least some of the documents are available by HTTP, so I'll try to use `self.session` when I get around to building that. Those should be even better for caching: the document at a particular URL should never change.

femtotrader:
I opened 2 enhancements upstream.

Here is an minimal example

```
import requests_ftp
import datetime
requests_ftp.monkeypatch_session()
import requests
import requests_cache
#session = requests.Session()
_URL = 'ftp://ftp.sec.gov/edgar/full-index/master.zip'
session = requests_cache.CachedSession(cache_name='cache', backend='sqlite', expire_after=datetime.timedelta(days=60))
response = session.get(_URL)
content = response.content
content = "{0}".format(content.decode('UTF-8', 'ignore'))
with open('master.zip', 'w') as fd:
    fd.write(content)
```

but zip is broken because of encoding I think.

jtkiley:
Cool. Here's a reworked version of your example. The initial download is a zip file, so it needs to be handled. I'll take a look at integrating this when I have a chance.

```python
import requests_ftp
import datetime
requests_ftp.monkeypatch_session()
import requests
import requests_cache
from pandas.io.common import ZipFile
from pandas.compat import BytesIO


# session = requests.Session()

_URL = 'ftp://ftp.sec.gov/edgar/full-index/master.zip'

s = requests_cache.CachedSession(cache_name='cache', backend='sqlite',
                                 expire_after=datetime.timedelta(days=60))
response = s.get(_URL)
content = response.content

content = BytesIO(content)

with ZipFile(content, 'r') as zf:
    index_file = zf.open(zf.namelist()[0]).read().decode()

with open('master.txt', 'w') as fd:
    fd.write(index_file)

```

femtotrader:
Great! Thanks for fixing my minimal example

jtkiley:
Where do I add `requests_ftp` so that the CI build will have it? I think `.travis.yml` is the right file.

femtotrader:
It should be in `setup.py` see

```
INSTALL_REQUIRES = (
    ['pandas', 'requests', 'requests-file']
)
```

jtkiley:
It should pass tests once the Eurostat servers stop throwing errors.

@femtotrader Thanks for the help and pointer to `requests_ftp`.

femtotrader:
If `pandas-datareader` uses `requests_ftp` we should probably help this upstream project by providing unit tests and continuous integration.

jtkiley:
@femtotrader That's a good idea. It looks like they have some tests already, too.

I'd like to get this merged and then finish working on adding support for EDGAR daily indices, but I'm happy to help after.

femtotrader:
An issue to have CI enabled is opened here https://github.com/Lukasa/requests-ftp/issues/16

davidastephens:
This looks good to me.  

If you're done, can you squash and I will merge?

jtkiley:
Done!

davidastephens:
Thanks!

PR #146, opened 2015-12-13
Clean FutureWarning on index order()

gliptak:


gliptak:
This fails to compile with the older (still supported) Pandas versions ...

bashtage:
Yes, you need a try except AttributeError or a general compatibility function for `sort`/`sort_values` changes.

PR #157, opened 2016-01-05
Added webscraper for Thrift Savings Plan

e2thenegpii:
I added a webscraper to put TSP fund data into a dataframe.  In order to import pandas_datareader on my machine I had to put a requirement requests>=2.3.0

### DIFF
--- a/docs/source/remote_data.rst
+++ b/docs/source/remote_data.rst
@@ -30,6 +30,7 @@ Currently the following sources are supported:
     - :ref:`World Bank<remote_data.wb>`
     - :ref:`OECD<remote_data.oecd>`
     - :ref:`Eurostat<remote_data.eurostat>`
+    - :ref:`Thrift Savings Plan<remote_data.tsp>`
 
 It should be noted, that various sources support different kinds of data, so not all sources implement the same methods and the data elements returned might also differ.
 
@@ -422,3 +423,16 @@ reconnect after waiting a few minutes.
     import pandas_datareader.data as web
     ed = web.DataReader('daily', 'edgar-index', '1998-05-18', '1998-05-18')
     ed[:5]
+
+.. _remote_data.tsp:
+
+TSP Fund Data
+
+Download mutual fund index prices for the TSP.
+
+.. ipython:: python
+
+    import pandas_datareader.tsp as tsp
+    tspreader = tsp.TSPReader(start='2015-10-1', end='2015-12-31')
+    tspreader.read()
+
--- a/pandas_datareader/base.py
+++ b/pandas_datareader/base.py
@@ -91,13 +91,21 @@ class _BaseReader(object):
         Open url (and retry)
         """
         response = self._get_response(url, params=params)
+        text = self._sanitize_response(response)
         out = StringIO()
-        if isinstance(response.content, compat.binary_type):
-            out.write(bytes_to_str(response.content))
+        if isinstance(text, compat.binary_type):
+            out.write(bytes_to_str(text))
         else:
-            out.write(response.content)
+            out.write(text)
         out.seek(0)
         return out
+
+    @staticmethod
+    def _sanitize_response(response):
+        """
+        Hook to allow subclasses to clean up response data
+        """
+        return response.content
 
     def _get_response(self, url, params=None):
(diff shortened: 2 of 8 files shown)


davidastephens:
Thanks!

Looks like the data is somewhat unreliable (see Travis Errors - I re-ran them and got the same thing).  Maybe add a check to ensure the scraper got data and a couple retries.  You can then raise RemoteDataError and add a  
```
try:
    #test code here
except RemoteDataError as e: # pragma: no cover
    raise nose.SkipTest(e)
`````
In the test.

Also, can you add a note in the what's new?

**Comment by e2thenegpii** [comment]
I looked at why the test was failing, the CSV file that I was downloading
from TSP.gov was rather irregular regarding the placement of whitespace and
newlines.  Pandas 14.1 and older didn't handle the sloppy whitespace well
which caused the error.  I added a callback in the DataReader class that is
called when extracting the text from the response object.  This should
enable subclasses to do things like fix up sloppy responses if needed as is
the case for TSP.  The default operation is to return the response object
content.

I also added a what's new 2.3 document.

File: docs/source/remote_data.rst [hunk 1]
```diff
@@ -409,3 +410,17 @@ Company filing index from EDGAR (SEC).
...
+
+TSP Fund Data
+=============
+
+Download mutual fund index prices for the TSP.
+
+.. ipython:: python
+
+    import pandas_datareader.tsp as tsp
+    tspreader = tsp.TSPReader(start='2015-10-1',end='2015-12-31')
```
femtotrader: a space before `end` and after comma is missing

    tspreader = tsp.TSPReader(start='2015-10-1', end='2015-12-31')

see https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements

pandas_datareader/base.py
```diff
@@ -92,14 +92,22 @@ def _read_url_as_StringIO(self, url, params=None):
...
+        if isinstance(text, compat.binary_type):
+            out.write(bytes_to_str(text))
         else:
-            out.write(response.content)
+            out.write(text)
         out.seek(0)
         return out
 
+    @classmethod
+    def _sanitize_response(cls,response):
```
femtotrader:
```python
    def _sanitize_response(cls, response):
```

pandas_datareader/tests/test_tsp.py
```diff
@@ -0,0 +1,16 @@
...
+import datetime as dt
+
+import pandas_datareader.tsp as tsp
+
+class TestTSPFunds(tm.TestCase):
+    def test_get_allfunds(self):
+        tspdata = tsp.TSPReader(start='2015-11-2',end='2015-11-2').read()
+        
+        assert len(tspdata == 1)
+        assert round(tspdata['I Fund'][dt.date(2015,11,2)],5) == 25.0058
```
femtotrader: fix space after coma

You may probably use `pandas.util.testing.assert_almost_equal`

pandas_datareader/tsp.py
```diff
@@ -0,0 +1,63 @@
...
+        df = super(TSPReader, self).read()
+        df.columns = map(lambda x:x.strip(),df.columns)
+        return df
+
+    @property
+    def params(self):
+        return {'startdate':self.start.strftime('%m/%d/%Y'),'enddate':self.end.strftime('%m/%d/%Y'),
+                'fundgroup':self.symbols,'whichButton':'CSV'}
+
+    @classmethod
```
davidastephens: Why is this a classmethod?

e2thenegpii: The _sanitize_response method shouldn't be changing the state of the reader
object but it is providing a service for the object.  It should only be
operating on the content of the response object.

davidastephens: Makes sense, but in that case I would use a staticmethod.

http://pythoncentral.io/difference-between-staticmethod-and-classmethod-in-python/

davidastephens:
Thanks - can you squash this and we can merge?

davidastephens:
I think this got squashed before the lint PR - can you rebase so we can merge?

Thanks.

e2thenegpii:
Everything should be set to go.  Thanks for your help.

davidastephens:
Thanks!

PR #159, opened 2016-01-24
Added retrieval of EDGAR daily indices.

### LINKED ISSUE #147: ENH: Add support for EDGAR daily indices.
EDGAR publishes a "full" index, that is an index of all filings in the current quarter, that `pandas-datareader` now supports. Once the quarter ends, its indices are moved to a folder and presented as daily indices. So, this enhancement will support pulling together a historical index for those of us who would like to use historical filings. I'm currently working on building this.

At the moment, I'm not planning on building in document retrieval directly, but I do want to make that reasonably easy. In my mind, a workflow would look like this:

1. Use `pandas-datareader` to pull the EDGAR index for the time period of interest. (This is where `pandas-datareader`'s role ends.)
2. Use a list of CIK identifiers, particular filing types, or another dataset (via merge) to filter down the list of documents that you'd like to retrieve.
3. Use `wget`, `curl`, or whatever to pull the documents.

When I say make it easy, I mean that I'm doing things like making the filename paths consistent in the returned index so that you can just concatenate the server name and the path to have a full link. In older data, the directory paths are missing a directory (presumably because they hadn't named it "EDGAR" yet).

My thinking on keeping document retrieval out is that these should be one-time pulls that we shouldn't cache, and the performance of a dedicated tool for pulling (potentially hundreds of thousands) documents should be far better than our readily-available options. Still, it should be trivial to create a download list from the index we return.

Any thoughts?

jtkiley:
Fixes #147.

### DIFF
--- a/docs/source/remote_data.rst
+++ b/docs/source/remote_data.rst
@@ -404,8 +404,21 @@ EDGAR Index
 
 Company filing index from EDGAR (SEC).
 
+The daily indices get large quickly (i.e. the set of daily indices from 1994
+to 2015 is 1.5GB), and the FTP server will close the connection past some
+downloading threshold . In testing, pulling one year at a time works well.
+If the FTP server starts refusing your connections, you should be able to
+reconnect after waiting a few minutes.
+
+
 .. ipython:: python
 
     import pandas_datareader.data as web
     ed = web.DataReader('full', 'edgar-index')
+    ed[:5]
+
+.. ipython:: python
+
+    import pandas_datareader.data as web
+    ed = web.DataReader('daily', 'edgar-index', '1998-05-18', '1998-05-18')
     ed[:5]
--- a/docs/source/whatsnew/v0.2.2.txt
+++ b/docs/source/whatsnew/v0.2.2.txt
@@ -20,6 +20,7 @@ New features
 
 - ``DataReader`` now supports dividend only pulls from Yahoo! Finance, see :ref:`here<remote_data.yahoo>` (:issue:`138`).
 - ``DataReader`` now supports SEC EDGAR full (current quarter) index retrieval, see :ref:`here<remote_data.edgar>` (:issue:`143`).
+- ``DataReader`` now supports SEC EDGAR daily (back to 7/1/1994) index retrieval, see :ref:`here<remote_data.edgar>` (:issue:`147`).
 
 .. _whatsnew_022.api_breaking:
 
--- a/pandas_datareader/compat/__init__.py
+++ b/pandas_datareader/compat/__init__.py
@@ -1,0 +1,1 @@
+from io import BytesIO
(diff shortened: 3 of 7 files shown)


jtkiley:
I'm not the only one getting these test errors and failures, right? I'm seeing it on Travis and locally on every test run.

```
ERROR: test_get_sts_cobp_a (pandas_datareader.tests.test_eurostat.TestEurostat)
FAIL: test_get_tourism (pandas_datareader.tests.test_oecd.TestOECD)
```

jtkiley:
Tests are all passing now except for the two unrelated ones I mentioned above and a timeout that's almost certainly a "too many connections" issue. They're (`ftp.sec.gov`) pretty quick to (briefly) block IPs for that. I tried to pile as many things as I could into a few tests, and I switched to `ftplib` to get each call to reuse one session. My real world use is working really well (i.e. full year pulls), because I'm not using a bunch of sessions.

@davidastephens: Let me know if there's anything I should fix before squashing. I'll probably add a little more to the docs about what works well.

@femtotrader: I ended up pulling out the `requests-ftp` stuff, because it doesn't reuse sessions. Even with small tests, I got a lot of errors from the server refusing to take additional connections. Let's keep an eye on it (or other options) to see if we have a caching alternative. This would be a good place for it, but it's thousands of small files rather than a large one.

jtkiley:
This is good to go unless you want me to use `try/except` in the tests to allow them to skip if they get `EOFError`. That's a sign that the FTP server has terminated the connection. We're getting it on Travis because of the number of testing instances and sessions in the tests, but it's not something I'm seeing in normal use. For some of my other commits (presumably when Travis was busier), they worked fine because the testing instances had staggered starts.

femtotrader:
After squashing it should be ok.
It sounds good to me.
@bashtage @davidastephens your opinion ?

davidastephens:
Strange, I thought I commented on this, but don't see it.

We should use a try: except and raise RemoteDataError where its possible to check that no data is received.  In the test, allow it to skip of there is a RemoteDataError.  We also need to do this for the Eurostat test that keeps failing.

jtkiley:
@davidastephens: Cool. I'm working on this right now. Do you want me to get those two Eurostat tests while I'm at it (in this PR)?

jtkiley:
Follow up: On the Eurostat and OECD tests that error/fail, I'm getting data when doing the same pulls that the tests are, but I think the labeling or data shape has changed in some way that isn't readily apparent, causing the tests (which are very specific) to fail.

I'd recommend skipping them in the short term. Let me know if you want me to do that in this PR.

davidastephens:
Thanks.  We should do the Eurostat ones on a separate PR.

jtkiley:
Squashed.

jtkiley:
Let me know if you want to merge #162 before or after a PR for skipping the failing Eurostat/OECD tests. It'll be a really simple PR, but it touches some of the same files that @femtotrader cleaned up a lot.

Once that's in, I'll add an issue for correcting the tests and share whatever I can find about how the data has changed.

davidastephens:
Thanks - I merged #162.

PR #165, opened 2016-02-05
Predictability and quality of life changes for EDGAR.

jtkiley:
Having pulled the whole series of daily indices (again), I made some slight changes to make the downloading process and returned data a little nicer to work with.

1. `dtype`s are now specified when reading the CSVs. This helps prevent type issues when reading batches and appending them to an `HDFStore`.
2. Dates are now converted to `Timestamp` using `to_datetime`.
3. Date and CIK are now used as index variables.
4. FTP sessions are now ended with `.close()`, as some seemingly-stalled downloads are only waiting on the server to respond to `QUIT`.

### DIFF
--- a/pandas_datareader/edgar.py
+++ b/pandas_datareader/edgar.py
@@ -20,6 +20,8 @@ _URL_DAILY = 'ftp://ftp.sec.gov/'
 _SEC_FTP = 'ftp.sec.gov'
 
 _COLUMNS = ['cik', 'company_name', 'form_type', 'date_filed', 'filename']
+_COLUMN_TYPES = {'cik': str, 'company_name': str, 'form_type': str,
+                 'date_filed': str, 'filename': str}
 _DIVIDER = re.compile('--------------')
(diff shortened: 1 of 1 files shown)


pandas_datareader/edgar.py
```diff
@@ -20,6 +20,8 @@
 _SEC_FTP = 'ftp.sec.gov'
 
 _COLUMNS = ['cik', 'company_name', 'form_type', 'date_filed', 'filename']
```
davidastephens: Do you still need this?

jtkiley: I do. `_COLUMNS` and `_COLUMN_TYPE` look similar and are used in `read_csv()`, but the first is used for naming columns and the second for `dtype`. I need to name the columns because the cleanest way of removing the varying lines of metadata (across small format changes over time) at the top of each csv also removes the header row.

davidastephens: Great, thanks.  I'll merge now.

Issue #173, opened 2016-02-23
REL: 0.3.0


davidastephens:
- [ ] Move tests for Python 2.6 and 3.3 to allowed failures
- [ ] Remove Python 2.6 and 3.3 workarounds

femtotrader:
Maybe we should put in doc a table with current status of each data source.
We should probably also define maintainer(s) per data source

|Data source name|Status|Maintainers|
| ----------- |------ | -----|
|yahoo|||
|yahoo-actions|||
|yahoo-dividends|||
|yahoo-dividends|||
|google|||
|fred|||
|famafrench|||
|oecd|||
|eurostat|||
|edgar-index|||
|yahoo-options|Broken #212||
|google-options|||

femtotrader:
Any news about this ?

@sinhrks @davidastephens

sinhrks:
+1 for releasing 0.3, adding pandas 0.19.0 support.

davidastephens:
Any objections to a new release?  A lot of things are broken with the previous release, so people can't use pip.  I'm happy to deal with putting it together.

gfyoung:
@jreback : `0.3.0` was released back in January.  Not sure why this is still open.

PR #174, opened 2016-02-23
BLD: Move python 2.6 & 3.3 to allowed failures

### LINKED ISSUE #160: Drop support for Python 2.6 and 3.3 (to match pandas 0.18.0)
As a point for discussion, we should consider dropping support for Python 2.6 and 3.3 to match the recent changes in pandas for 0.18.0. Among other things, it looks like it may impact our `pandas.compat` imports. If we decide to follow, we may also want to look at the associated pandas versions that we test against on 2.6 and 3.3.

See:
https://github.com/pydata/pandas/issues/7718
https://github.com/pydata/pandas/pull/11988
https://github.com/pydata/pandas/issues/11273

davidastephens:
#173 
closes #160


gliptak:
Why don't we remove ```2.6``` and ```3.3``` from the build altogether? Thanks

PR #181, opened 2016-02-27
Set chopdata to private in Yahoo options

gliptak:


gliptak:
@femtotrader This fails on -0.0% decrease of coverage ...

femtotrader:
No problem... let's merge it

PR #182, opened 2016-02-27
COMPAT: Do not pass iterator as index

sinhrks:
Fixed Travis Error in ``test_get_quote_stringlist``, which raises ``TypeError: data argument can't be an iterator``.

This makes CI pass except py2.6 / pandas 0.13.1 which looks to have a different issue.


sinhrks:
Simply skipped the problematic test ``test_month_year`` in python 2.6 (#169).

femtotrader:
Thanks @sinhrks

PR #187, opened 2016-03-01
Add pyi version and downloads badges

gliptak:


femtotrader:
Good idea!

Thanlks @gliptak

PR #189, opened 2016-03-05
CLN: Cleanup option reader

sinhrks:
Cleanup and added some tests.

### DIFF
--- a/pandas_datareader/base.py
+++ b/pandas_datareader/base.py
@@ -223,3 +223,61 @@ def _in_chunks(seq, size):
     Return sequence in 'chunks' of size defined by size
     """
     return (seq[pos:pos + size] for pos in range(0, len(seq), size))
+
+
+class _OptionBaseReader(_BaseReader):
+
+    def __init__(self, symbol, session=None):
+        """ Instantiates options_data with a ticker saved as symbol """
+        self.symbol = symbol.upper()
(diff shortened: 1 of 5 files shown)


gliptak:
@sinhrks I would like to propose simplifying the Options approach. The steps to obtain options data would be:

1. Get list of expiry dates (varies with weekly vs monthly only)
2. Get all options data for a particular expiry
3. Filter for call/put, strikes near current price, aggregate for all expiries for current month are to happen in calling code

PS Some of the current approach (in particular call/put separation) seem to have been driven by the Yahoo! web representation ...

sinhrks:
Thanks. Because I don't familiar with Yahoo/google quotes API, could you do it separately?

gliptak:
@sinhrks I'm not sure what you refer above as "separately" (in a separate pull request?) What I'm describing above, is a proposed cleanup of the API (really leaving two methods, get ```expiry_dates``` and ```get_options_data``` for a particular expiry date) working the same for both Yahoo and Google.

sinhrks:
Yes, separate PR. I'm not quite understood the suggested change.

gliptak:
Sure. I will wait until this one is committed.

sinhrks:
If there is no comments, will merge within few days.

femtotrader:
+1 let's merge it! thanks @gliptak

PR #199, opened 2016-04-27
ENH: add freq parameter to _BaseReader

### LINKED ISSUE #198: World Bank datareader cannot retrive monthly data
I am trying to extract data from the World Bank and have been successful with the following code via `pandas.io`:
```python
from pandas.io import wb
df_wb = wb.download(indicator=['GOLD','SILVER','COPPER','ZINC','LEAD','NICKEL', 
                             'ALUMINUM'],country='ALL',start='1980M01',end='2016M12')
```
This works and returns the data as desired even though the `start` and `end` expect `int` inputs. This is likely to be as a result of no checking on the `start` and `end` inputs.

The same code above employing `pandas-datareader` fails. Given there is no facility in `pandas-datareader` to specify the frequency of the data requested from the World Bank the data returned when supplying a `datetime` object or `int` is limited to an annual frequency.

To facilitate the returning of monthly data the call signature for `_BaseReader` (in `base.py`) would need to be updated to include a parameter to specify the frequency, for instance `...,freq='M',...` consistent with `pandas` naming, and for this to be taken into account when building the url for the World Bank database (in `wb.py`).

```python
@property
def params(self):
    return {'date': '{0}:{1}'.format(self.start.year, self.end.year),
        'per_page': 25000, 'format': 'json'}
```
becomes
```python
@property
def params(self):
   if self.freq = 'M'
        return {'date': '{0}M{1:02d}:{2}M{3:02d}'.format(self.start.year, self.start.month, 
            self.end.year self.end.month),'per_page': 25000, 'format': 'json'}
    return {'date': '{0}:{1}'.format(self.start.year, self.end.year),
        'per_page': 25000, 'format': 'json'}
```

kurtforrester:
added `freq` paramter `_BaseReader` and updated `def params` to accomodate monthly frequency data from the World Bank database.

closes #198

### DIFF
--- a/pandas_datareader/base.py
+++ b/pandas_datareader/base.py
@@ -42,13 +42,14 @@ class _BaseReader(object):
     _chunk_size = 1024 * 1024
     _format = 'string'
 
-    def __init__(self, symbols, start=None, end=None,
+    def __init__(self, symbols, start=None, end=None, freq=None,
                  retry_count=3, pause=0.1, session=None):
         self.symbols = symbols
 
         start, end = self._sanitize_dates(start, end)
         self.start = start
         self.end = end
+        self.freq = freq
 
         if not isinstance(retry_count, int) or retry_count < 0:
             raise ValueError("'retry_count' must be integer larger than 0")
--- a/pandas_datareader/base.py
+++ b/pandas_datareader/base.py
@@ -42,14 +42,13 @@ class _BaseReader(object):
     _chunk_size = 1024 * 1024
     _format = 'string'
 
-    def __init__(self, symbols, start=None, end=None, freq=None,
+    def __init__(self, symbols, start=None, end=None,
                  retry_count=3, pause=0.1, session=None):
         self.symbols = symbols
 
         start, end = self._sanitize_dates(start, end)
         self.start = start
         self.end = end
-        self.freq = freq
 
         if not isinstance(retry_count, int) or retry_count < 0:
             raise ValueError("'retry_count' must be integer larger than 0")
--- a/pandas_datareader/base.py
+++ b/pandas_datareader/base.py
@@ -149,16 +149,18 @@ class _BaseReader(object):
         if is_number(start):
             # regard int as year
             start = dt.datetime(start, 1, 1)
-        start = to_datetime(start)
 
         if is_number(end):
             end = dt.datetime(end, 1, 1)
-        end = to_datetime(end)
 
         if start is None:
             start = dt.datetime(2010, 1, 1)
         if end is None:
             end = dt.datetime.today()
+
+        start = to_datetime(start)
+        end = to_datetime(end)
+
         return start, end
 
 
--- a/pandas_datareader/wb.py
+++ b/pandas_datareader/wb.py
@@ -117,7 +117,7 @@ class WorldBankReader(_BaseReader):
     _format = 'json'
 
     def __init__(self, symbols=None, countries=None,
-                 start=None, end=None,
+                 start=None, end=None, freq=None,
                  retry_count=3, pause=0.001, session=None, errors='warn'):
 
         if symbols is None:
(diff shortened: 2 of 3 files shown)


sinhrks:
Thanks, can you add tests?

pandas_datareader/base.py
```diff
@@ -42,13 +42,14 @@ class _BaseReader(object):
     _chunk_size = 1024 * 1024
     _format = 'string'
 
-    def __init__(self, symbols, start=None, end=None,
+    def __init__(self, symbols, start=None, end=None, freq=None,
```
sinhrks: raise `ValueError` if freq is not `None` or `"M"`

kurtforrester:
I don't know how to write tests but I will learn. Can you (@sinhrks) point me to where they need to be, what one looks like and I will work out what I have to do?

pandas_datareader/base.py
kurtforrester: It may be that the base datareader should accept all of the valid pandas/datetime freq strings ('H', 'M', 'Q', 'A',...) and the specific modules for wb or oecd etc should then determine what is suitable for the particular data source? For instance World Bank offers monthly data, quarterly data, and annual data for the various data sources.

sinhrks:
You can find some examples here.

- https://github.com/pydata/pandas-datareader/blob/master/pandas_datareader/tests/test_wb.py

pandas_datareader/base.py
sinhrks: No need to care values which is not implemented ATM from user's point of view. Also, pls move `freq` to `WorldBankReader` as other readers don't use it.

NOTE: Some other readers have `interval` kw which also accept frequency-like. It may better to use the same word but not sure as I'm not native English speaker.

kurtforrester:
I have reversed changes I made to the base data reader class and bundled it all in the world bank reader. I have added two test, one for monthly data and one for quarterly. Both currently pass on my machine (Python 3.5.1 from Anaconda 2.5.0 (64 bit) - Windows). Let me know if there is anything else needing attention.

pandas_datareader/wb.py
```diff
@@ -154,8 +160,18 @@ def url(self):
 
     @property
     def params(self):
-        return {'date': '{0}:{1}'.format(self.start.year, self.end.year),
-                'per_page': 25000, 'format': 'json'}
+        if self.freq == 'M':
+            return {'date': '{0}M{1:02d}:{2}M{3:02d}'.format(self.start.year,
+                    self.start.month, self.end.year, self.end.month),
+                    'per_page': 25000, 'format': 'json'}
+        if self.freq == 'Q':
```
sinhrks: pls use ``if-elif-else``

pandas_datareader/wb.py
```diff
@@ -154,8 +160,18 @@ def url(self):
...
+            return {'date': '{0}Q{1}:{2}Q{3}'.format(self.start.year,
+                    divmod(self.start.month-1,3)[0]+1, self.end.year,
```
sinhrks: Can you change ``_sanitize_dates`` to always return ``Timestamp``, and can use ``self.start.quarter``?

- https://github.com/pydata/pandas-datareader/blob/master/pandas_datareader/base.py#L143

femtotrader:
I like `freq` parameter name as it's similar to `pandas.date_range` parameter http://pandas.pydata.org/pandas-docs/stable/generated/pandas.date_range.html

but I'd convert `freq` (string) to Pandas DateOffset
http://stackoverflow.com/questions/27368265/convert-freq-string-to-dateoffset-in-pandas

what is your opinion about this @sinhrks ?

sinhrks:
No opinion for `DateOffset`, because no benefit from conversion ATM.

sinhrks:
Failire looks to be unrelated from the change. Let me check...

davidandreoletti:
@sinhrks @femtotrader When is this PR going to be merged ? I actually need it :)

sinhrks:
Sorry, should have noticed that test failure has been fixed in #204.

@kurtforrester can u rebase?

bashtage:
@kurtforrester is this still needed/useful?

PR #201, opened 2016-05-03
Fix Yahoo Options expiry date

bearnshaw:


femtotrader:
Thanks @bearnshaw 
See https://github.com/pydata/pandas-datareader/pull/203

Issue #191, opened 2016-03-15
Options to_*() is exporting the expiry date incorrectly


vishalapte:
```
In [1]: from pandas_datareader import data as pddata

In [2]: foo = pddata.Options('aapl', 'yahoo')

In [3]: bar = foo.get_all_data()

In [4]: bar.to_csv()

'Strike,Expiry,Type,Symbol,Last,Bid,Ask,Chg,PctChg,Vol,Open_Int,IV,Root,IsNonstandard,Underlying,Underlying_Price,Quote_Time
10.0,2017-06-16,call,AAPL160617C00010000,0.0,91.0,91.45,0.0,0.0,0,0,0.0,AAPL,False,AAPL,102.52,2016-03-15 16:00:00
10.0,2017-06-16,put,AAPL160617P00010000,0.0,0.0,0.03,0.0,0.0,0,0,1.5938,AAPL,False,AAPL,102.52,2016-03-15 16:00:00
12.5,2017-06-16,call,AAPL160617C00012500,0.0,88.5,88.95,0.0,0.0,0,0,0.0,AAPL,False,AAPL,102.52,2016-03-15 16:00:00
...
```

**ERROR**: 20DD-MM-YY is erroneously being exported as Expiry instead of 20YY-MM-DD. The error is consistent across to_csv(), to_json, to_dict(), etc.

```
In [1]: import pandas_datareader

In [2]: pandas_datareader.version
Out[2]: '0.2.1'
```

davidastephens:
This looks to be fixed with #244.  Please reopen if not.

PR #203, opened 2016-05-18
Yahoo Options Added format specifier for date parsing of options.

adamjoshuagray:
I happened to notice that dates were getting parsed in the reverse order for me. I suspect that was because of a locale issue and so I added an explicit format specifier for the dates which alleviates that problem.


femtotrader:
Seems similar to https://github.com/pydata/pandas-datareader/pull/203

femtotrader:
Closes https://github.com/pydata/pandas-datareader/issues/191

femtotrader:
Thanks @adamjoshuagray

Issue #149, opened 2015-12-17
Eurostat: Not working for a number of data series


esvhd:
Hi guys,

I've been trying to get data out of Eurostat, however, I have not had much luck with data series other than the one given in the docs [here](http://pandas-datareader.readthedocs.org/en/latest/remote_data.html#remote-data-eurostat).

So if I try to get the HICP data, I get the following error. Does anyone have any suggestion on how to fix it?

Thanks!
Will

```
import pandas_datareader.data as web
    
HICP_YOY = 'prc_hicp_manr'
# tried 'gov_10q_ggdebt' as well, same error.
hicp = web.DataReader(HICP_YOY, 'eurostat')
    
Output:
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-11-e31c150c6129> in <module>()
----> 1 hicp = web.DataReader('gov_10q_ggdebt', 'eurostat')

/Users/zwl/anaconda/lib/python3.4/site-packages/pandas_datareader/data.py in DataReader(name, data_source, start, end, retry_count, pause, session)
    122         return EurostatReader(symbols=name, start=start, end=end,
    123                               retry_count=retry_count, pause=pause,
--> 124                               session=session).read()
    125     else:
    126         raise NotImplementedError(

/Users/zwl/anaconda/lib/python3.4/site-packages/pandas_datareader/base.py in read(self)
     73     def read(self):
     74         """ read data """
---> 75         return self._read_one_data(self.url, self.params)
     76 
     77     def _read_one_data(self, url, params):

/Users/zwl/anaconda/lib/python3.4/site-packages/pandas_datareader/eurostat.py in _read_one_data(self, url, params)
     33 
     34         resp = self._get_response(url)
---> 35         data = read_sdmx(resp.content, dsd=dsd)
     36 
     37         try:

/Users/zwl/anaconda/lib/python3.4/site-packages/pandas_datareader/io/sdmx.py in read_sdmx(path_or_buf, dtype, dsd)
     51     root = ET.fromstring(xdata)
     52 
---> 53     structure = _get_child(root, _MESSAGE + 'Structure')
     54     idx_name = structure.get('dimensionAtObservation')
     55     dataset = _get_child(root, _DATASET)

/Users/zwl/anaconda/lib/python3.4/site-packages/pandas_datareader/io/sdmx.py in _get_child(element, key)
    143         return elements[0]
    144     elif len(elements) == 0:
[...]

PR #204, opened 2016-06-11
TST: Fix broken tests

sinhrks:
Fixed broken tests and updated pandas version tested.

- https://travis-ci.org/pydata/pandas-datareader/builds/125410265

An Eurostat test is temporary disabled because now it meets the same issue as #149.


sinhrks:
Merging within few days if no comments.

PR #209, opened 2016-06-25
Permute tests and PEP8

femtotrader:
I think it will be better to first see if unit tests are passing before checking PEP8 conformance.


femtotrader:
No comment except @sinhrks 👍 
So I'm merging

PR #245, opened 2016-09-28
Adds support for Enigma datasets

trevorprater:
Usage:
```
df = pdr.get_data_enigma('enigma.inspections.restaurants.fl')
```

If `ENIGMA_API_KEY` does not exist in your env, it can be supplied as the second arg:

```
df = pdr.get_data_enigma('enigma.inspections.restaurants.fl', 'ARIAMFHKJMISF38UT')
```

### DIFF
--- a/pandas_datareader/enigma.py
+++ b/pandas_datareader/enigma.py
@@ -1,0 +1,107 @@
+import zlib
+import json
+import os
+import sys
+import time
+from pandas.compat import StringIO
+
+from pandas import DataFrame
+import pandas.compat as compat
+import pandas as pd
+import requests
+
+from pandas_datareader.base import _BaseReader
+
+
+class EnigmaReader(_BaseReader):
+    """
+    Collects Enigma data located at the specified datapath and returns a pandas DataFrame.
+
+    Usage (high-level):
+    ```
+        import pandas_datareader as pdr
+        df = pdr.get_data_enigma('enigma.inspections.restaurants.fl')
+
+        #in the event that ENIGMA_API_KEY does not exist in your env, it can be supplied as the second arg:
+        df = prd.get_data_enigma('enigma.inspections.restaurants.fl', 'ARIAMFHKJMISF38UT')
+    ```
+
+    Usage:
+    ```
+        df = EnigmaReader(datapath='enigma.inspections.restaurants.fl', api_key='ARIAMFHKJMISF38UT').read()
+    ```
+    """
+
+    def __init__(self,
+                 datapath=None,
+                 api_key=None,
+                 retry_count=5,
+                 pause=0.250,
+                 session=None):
+
+        super(EnigmaReader, self).__init__(symbols=[],
+                                           retry_count=retry_count,
+                                           pause=pause)
+        if api_key == None:
+            self._api_key = os.getenv('ENIGMA_API_KEY')
+            if self._api_key == None:
+                raise ValueError(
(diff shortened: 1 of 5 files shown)


femtotrader:
Thanks @trevorprater

Could you add `DataReader` as an entry point also ?
Ideally most DataReaders use it an entry point.

A simple test will be nice also.

`StringIO` can be import using

    from pandas.compat import StringIO

pandas_datareader/enigma.py
```diff
@@ -0,0 +1,109 @@
...
+        return 'export_url'
+
+    
+    @property
+    def _head_key(self):
+        return 'head_url'
+
+    
+    def _request(self, url):
+        resp = requests.get(url)
```
femtotrader: You should use

    resp = self.session.get(url)

so query can be cached.

trevorprater:
Thank you for the code review, @femtotrader! I have made the suggested changes. The only thing left is for me to write some tests, I think.

trevorprater:
Tests have been added and I believe all requested changes have been made. Thank you, @femtotrader!

femtotrader:
Thanks @trevorprater 

but I noticed you add directly API key `yddA...Nnm` into test.
That's not a very good idea because anyone looking at the code could use this API key.
That's probably not an issue for Enigma (maybe you restricted this API key) but if other users/developers want to code their own DataReader and they have a look at this code they could make a similar security error.
I suggest to use environment variable for this.
See https://docs.travis-ci.com/user/environment-variables/

Please also squash commits
http://pandas.pydata.org/pandas-docs/stable/contributing.html#contributing-your-changes-to-pandas

I suggest a commit message like "ENH: Adds support for Enigma datasets"

trevorprater:
Thank you for the additional feedback, @femtotrader.

I have added the support for an environment variable `ENIGMA_API_KEY` to the tests. Apologies for adding it directly to the code. I assumed it would be necessary for tests to pass. I agree that it definitely makes more sense to use an env var.

I have also squashed the commits and updated the commit message.

- Trevor

femtotrader:
Thanks again @trevorprater 

Other contributions are welcome!

sinhrks:
I think it never succeeded on Travis because we don't set  ``ENIGMA_API_KEY``. can u fix?

- https://travis-ci.org/pydata/pandas-datareader/pull_requests

jreback:
i just turned on the check that requires travis passing before merging

sinhrks:
I've prepared #250 to fix other broken tests. For this PR, we should decide either:

- revert this once, until the test is being ready
- make a separate issue and leave it using ``SkipTest``

femtotrader:
I'm for `SkipTest` use.

sinhrks:
How about:

- Merge #250 and prepare v0.3.0 release. 
- If @trevorprater can update the tests until the release, it should be in 0.3.0. Otherwise, postpone the functionality to the next release.

femtotrader:
#250 is merged.

I don't understand what you would like in tests ?
Using `assertRaises` like https://github.com/pydata/pandas-datareader/blob/ca4054c50a6e9268496cdf726d32d7fc96c96811/pandas_datareader/tests/test_base.py#L8 instead of `_exception`
or is something else required ?

What can be the side effects of using `SkipTest` ?

sinhrks:
What I meant is to make Enigma tests be runnable on Travis to guarantee the function. Currently no others run its tests, correct?

PR #254, opened 2016-11-06
ENH: Add support for Nasdaq symbol list

robottwo:
Nasdaq provides a symbol list with all US equities,
listed on the Nasdaq and other exchanges on their
FTP site. This patch adds support for grabbing
that list.

The list isn't very larger (< 1 MB), so we cache it
to RAM within the process the first time we download it.


femtotrader:
Thanks @robottwo for this contribution.

There is unit tests, documentation.
flake8 is happy...

I'm ok for merging but I will wait about one week before doing this to allow other people to give their feedback about this PR.

Thanks again

femtotrader:
Thanks again @robottwo

PR #255, opened 2016-11-07
Add OANDA Rest Historical Instrument Reader

davidandreoletti:


davidandreoletti:
Still some work to do on this PR (i.e dependency oandapy fixup). Deleting it for now.

femtotrader:
Thanks @davidandreoletti 

But you could simply give this PR a title starting with WIP and still push there.

In order to provide a better user API experience `granularity` could be a Pandas offset shortcut string (like '5Min', '5H' ... in `resample` function ).

And it could be convert to a string compatible with Oanda API.

We could have a parameter named `freq` for `DataReader` function (it's role is quite similar to `pandas.date_range` parameter)

See https://github.com/pydata/pandas-datareader/pull/199

I'd convert `freq` (string) to Pandas DateOffset
http://stackoverflow.com/questions/27368265/convert-freq-string-to-dateoffset-in-pandas

and have a dict to convert Pandas DateOffset to Oanda granularity string.

Ask me if you don't understand what I mean (because I'm not english native speaker) so I may miss something.

femtotrader:
superseed by https://github.com/pydata/pandas-datareader/pull/257

PR #264, opened 2016-12-06
ENH: _BaseReader can be used as context guard.

nme01:
Now if someone creates a lot of requests, he can use Readers using the `with` statement so that the session is correctly closed, e.g.:
```python
with YahooDailyReader(symbols, start=start, end=end) as reader:
    return reader.read()
```

### DIFF
--- a/pandas_datareader/base.py
+++ b/pandas_datareader/base.py
@@ -55,6 +55,12 @@ class _BaseReader(object):
         self.timeout = timeout
         self.session = _init_session(session, retry_count)
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exception_type, exception_value, exception_traceback):
+        self.session.close()
+
     @property
     def url(self):
         # must be overridden in subclass
--- a/pandas_datareader/data.py
+++ b/pandas_datareader/data.py
@@ -25,35 +25,43 @@ from pandas_datareader.nasdaq_trader import get_nasdaq_symbols
 
 
 def get_data_fred(*args, **kwargs):
-    return FredReader(*args, **kwargs).read()
+    with FredReader(*args, **kwargs) as reader:
+        return reader.read()
 
 
 def get_data_famafrench(*args, **kwargs):
-    return FamaFrenchReader(*args, **kwargs).read()
+    with FamaFrenchReader(*args, **kwargs) as reader:
+        return reader.read()
 
 
 def get_data_google(*args, **kwargs):
-    return GoogleDailyReader(*args, **kwargs).read()
+    with GoogleDailyReader(*args, **kwargs) as reader:
+        return reader.read()
 
 
 def get_data_yahoo(*args, **kwargs):
-    return YahooDailyReader(*args, **kwargs).read()
+    with YahooDailyReader(*args, **kwargs) as reader:
+        return reader.read()
 
 
 def get_data_enigma(*args, **kwargs):
-    return EnigmaReader(*args, **kwargs).read()
+    with EnigmaReader(*args, **kwargs) as reader:
+        return reader.read()
 
 
 def get_data_yahoo_actions(*args, **kwargs):
-    return YahooActionReader(*args, **kwargs).read()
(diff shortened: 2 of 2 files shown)


femtotrader:
Maybe you should add a test case (or modify an existing one) to ensure that this new feature works as expected.

nme01:
I'm not exactly sure how this should be done. I would need to start a connection with something (e.g. using session's mock). However `_BaseReader` is incapable of reading anything from a session (even if I'd used session's mock) since its `url` property is abstract (raises `NotImplementedError`). Could you give me a hand with this?

bashtage:
You could test using anything that derives from _BaseReader that should
inherit the context manager.

bashtage:
Is there a good reason that `read()` doesn't always close?  If it was implemented this way there would not be a need for a CM.

femtotrader:
Maybe because you can use same `session` for several readings.

bashtage:
That was my question.  Is this a realistic use case and is there any gain in allowing session reuse?

femtotrader:
I wonder why we should recreate an other `Session` (or `CachedSession`) instance when parameters to build them are the same?  Is there any gain in instantiating again a `Session` (or `CachedSession`?
Most of us never close `session` when using for example IPython. So my question would be why closing `session` would be required. I think it's not necessary to do it (we never did it previously) and it didn't hurt. So the idea of using a context manager is nice as it doesn't urge people to use it.
But I won't use this CM is `get_data_fred`, `get_data_famafrench` ... because it will close session by default.
I will only use this CM in one test of a DataReader that derives from `_BaseReader`

nme01:
However when you use `get_data_fred`, `get_data_famafrench`, `...` you create a new session on each call. There is no option to pass an already opened session. It doesn't even return the created session. I created the pull request, because we got a lot of `ResourceWarnings` when we ran our tests with `unittests` each time the read failed (e.g. because it couldn't reach YahooFinance). It should be ensured that the session is properly closed in these cases.

femtotrader:
Quote from: @nme01 

    However when you use get_data_fred, get_data_famafrench, ... you create a new session on each call. There is no option to pass an already opened session. It doesn't even return the created session. I created the pull request, because we got a lot of ResourceWarnings when we ran our tests with unittests each time the read failed (e.g. because it couldn't reach YahooFinance). It should be ensured that the session is properly closed in these cases.

It seems that `get_data_fred`, `get_data_famafrench` accept keyword args and `session` is one.
You don't need to return the "created session", session is an object on which you can keep a reference.

nme01:
I realised that right after I posted my comment. That's why I've deleted it.

pandas_datareader/base.py
```diff
@@ -55,6 +55,12 @@ def __init__(self, symbols, start=None, end=None,
         self.timeout = timeout
         self.session = _init_session(session, retry_count)
```
bashtage: You should keep track of whether it was a user session or a new session.

pandas_datareader/base.py
```diff
@@ -55,6 +55,12 @@ def __init__(self, symbols, start=None, end=None,
...
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exception_type, exception_value, exception_traceback):
+        self.session.close()
```
bashtage: Only close if it was a new session.

pandas_datareader/data.py
```diff
@@ -25,35 +25,43 @@
...
 
 def get_data_google(*args, **kwargs):
-    return GoogleDailyReader(*args, **kwargs).read()
+    with GoogleDailyReader(*args, **kwargs) as reader:
+        return reader.read()
 
 
 def get_data_yahoo(*args, **kwargs):
-    return YahooDailyReader(*args, **kwargs).read()
+    with YahooDailyReader(*args, **kwargs) as reader:
```
bashtage: Don't need to test all readers.  Lets  stick to Fred and FamaFrench since these are likely to work.

bashtage:
I'm going to close this since I don't see a compelling case.

PR #275, opened 2017-01-14
REL: Bump version to fix pypi upload

davidastephens:


bashtage:
Is this PEP-compliant versioning?  FWIW I think it is possible to delete releases through the PyPi website and then to repost (even though PyPI makes it hard).

davidastephens:
pypi changed it to 0.3.1, I'll update the docs/release tag.  I've tried to find a way to delete them and repost and it doesn't seem possible.  My bad, I forgot to fetch upstream and missed a commit.

davidastephens:
Never mind, I think I found a way to remove and re-upload.  What do you think, delete 0.3.0 and re-upload or just go straight to 0.3.1?

bashtage:
I would use 0.3.0 since there has been virtually no time since initial release.  0.3.1 is safer though in case someone somewhere installed 0.3.0.

davidastephens:
Pypi won't let me upload a file with the same name as the previous one, so we have to name it differently.  I'm thinking v0.3.0.post, per PEP 440:

Some projects use post-releases to address minor errors in a final release that do not affect the distributed software (for example, correcting an error in the release notes).

If used as part of a project's development cycle, these post-releases are indicated by including a post-release segment in the version identifier

PR #286, opened 2017-03-15
Update quotes.py

internalconsistence:
Hi, I would like to add the price to book value in the quote. Thanks!


femtotrader:
Thanks @internalconsistence but please open an issue.
Explain what is missing and after open a PR.

bashtage:
Going to close since poorly documented, no supporting motivation and Yahoo is generally problematic now.

PR #304, opened 2017-05-05
COMPAT for pandas 0.20.0

jreback:


jreback:
so this will all pdr to be imported; we had removed ``PandasError``. However there are still lots of testing stuff that is using 'internal' pandas things. So this needs to change.

@gfyoung any interest in fixing?

jreback:
cc @davidastephens

gfyoung:
@jreback : I don't fully understand what you wrote.  Could you clarify?

jreback:
@gfyoung the test suite here is a) nose, b) uses lots of things we removed from pandas.util.testing. So best case would be to simply convert this to use pytest and use the public things in pandas.testing.

gfyoung:
@jreback : Ah, I see.  I can have a look.

bashtage:
LGTM

gfyoung:
@jreback : I think it can be done, but I'm going to wait until this PR gets merged

jreback:
ok, I kind of wanted to have actual tests. But yes this *does* fix the issues.

jreback:
@gfyoung go to it!

gfyoung:
@jreback : Travis seems to disagree that all compat issues have been addressed.

gfyoung:
Actually, I see it's probably because `api.types` does not exist in older version of `pandas`.  Also, any reason that none of the tests install `pandas` from `master` ?

gfyoung:
I'm putting up PR to immediately patch `api.types` failure.

jreback:
@gfyoung the import works (which was the initial issue). now none of the tests work because
a) no testing on travis of 0.20.1 (don't need master, this is enough)
b) tm.TestCase is relied upon.
c) api.types.

jreback:
FYI this still uses nose. I think easiest would be to simply change to pytest. (or can just remove tm.TestCase and fix things too)

gfyoung:
@jreback : No, I believe that the test works when the CORRECT version of `pandas` is installed.  However, I don't see why not just patch the `api.types` failure (condition on `pandas` version) for the moment.

gfyoung:
> no testing on travis of 0.20.1 (don't need master, this is enough)

Why not just add `master` to allowed failures?  Couldn't hurt, could it?

> tm.TestCase is relied upon

Well, that will be taken care of when I do the `pytest` conversion.

jreback:
> Why not just add master to allowed failures? Couldn't hurt, could it?

sure, but have to install from daily wheels (which can do), but easier just to install latest :>

can do it in steps.

gfyoung:
> sure, but have to install from daily wheels (which can do), but easier just to install latest

Fair enough.  I would prefer to install from daily wheels in the end, though I'm not sure how that is done.

> can do it in steps.

Agreed.

jreback:
https://github.com/pandas-dev/pandas/blob/master/ci/requirements-3.6_NUMPY_DEV.build.sh
except install pandas (and just install numpy first, otherwise I think it will attempt to build it which is bad).

mrocklin:
When is this likely to be released?  I'm running a tutorial on Wednesday that uses this library and want to pin versions.  Should I pin 0.20.1 or 0.19?

jreback:
https://github.com/pydata/pandas-datareader/issues/328

PR #309, opened 2017-05-07
COMPAT: Make is_number import backwards-compat

gfyoung:
Title is self-explanatory.

xref #304.

### DIFF
--- a/pandas_datareader/edgar.py
+++ b/pandas_datareader/edgar.py
@@ -3,16 +3,13 @@ import datetime as dt
 from ftplib import FTP
 import gzip
 
-from pandas import read_csv
-from pandas import DataFrame
-from pandas import to_datetime
 from zipfile import ZipFile
 from pandas.compat import StringIO
-from pandas.api.types import is_number
+from pandas import read_csv, DataFrame, to_datetime
 
 from pandas_datareader.base import _BaseReader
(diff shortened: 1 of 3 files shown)


pandas_datareader/edgar.py
```diff
@@ -1,19 +1,23 @@
...
 from pandas.compat import StringIO
-from pandas.api.types import is_number
+from distutils.version import LooseVersion
+from pandas import read_csv, DataFrame, to_datetime
 
 from pandas_datareader.base import _BaseReader
 from pandas_datareader.compat import BytesIO
 from pandas_datareader._utils import RemoteDataError
 
+if LooseVersion(pd.__version__) <= '0.19.2':
```
jreback: would put this in ``pandas_datareader/compat/__init__.py``

```
if LooseVersion(.....):
       from pandas.core import common as pdapi
else:
      from pandas.api import types as pdapi
```

then you can 

``from pandas_datareader.compat import pdapi``

gfyoung: That works too.

jreback: if directly import ``is_number`` in compat is ok too. this then isolates all version specific things there.

gfyoung: That works too.

jreback:
thanks!

PR #310, opened 2017-05-07
TST: Use pytest instead of nose

gfyoung:
Use `pytest` instead of `nose` per discussion in #304.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ install:
   - conda update -q conda
   # Useful for debugging any issues with conda
   - conda info -a
-  - conda create -q -n test-environment python=$PYTHON pandas=$PANDAS nose coverage setuptools html5lib lxml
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -30,7 +30,7 @@ install:
   - conda update -q conda
   # Useful for debugging any issues with conda
   - conda info -a
-  - conda create -q -n test-environment python=$PYTHON pandas=$PANDAS nose coverage setuptools html5lib lxml
+  - conda create -q -n test-environment python=$PYTHON pandas=$PANDAS nose coverage setuptools html5lib lxml pytest pytest-cov
```
jreback: remove nose

gfyoung: Not yet.  Got a bunch `nose.SkipTest` things to remove.  I just wanted to make sure that initial groundwork (i.e. using it) works.  Then we can remove `nose` things.

PR #319, opened 2017-05-09
TST: Remove self.assert* from testing

gfyoung:
Title is self-explanatory.

Remaining instances are those where an `assert*` function was defined directly in the class.


bashtage:
Move `TestCase` import location, or remove completely since not needed for pytest.

gfyoung:
Will be removed in follow-up.  We got rid of the `TestCase` class in `0.20.0`.

jreback:
thanks

PR #320, opened 2017-05-10
COMPAT, TST: Remove tm.TestCase from testing

gfyoung:
Title is self-explanatory.  Now all build machines should not have issues testing due to incompatibilities.


gfyoung:
I think the `test_eurostat.py` websites are down, so those tests might fail across the board.  I also got some failures on `0.20.0` beyond the `tm.TestCase` incompatibilities.  However, those could be fixed afterwards.

jreback:
thanks!

PR #325, opened 2017-05-11
MAINT: Move all tests under one directory

gfyoung:
Title is self-explanatory.  Also move `google` and `yahoo` tests into own directories to be consistent with file structure of the code-base.


jreback:
thanks!

PR #335, opened 2017-05-18
remove string ',' in quote price

silau2005:
replace(',','')
this will facilitate the conversion from string to float, and avoid exception


silau2005:
this string ',' removal is especially important for handling index price
e.g. INDEXHANGSENG:HSI
the original quote price is a string with ',' to show the thousand price point
which will cause error when converting to float

femtotrader:
`read_csv` have a parameter `thousands`
see http://pandas.pydata.org/pandas-docs/stable/generated/pandas.read_csv.html

femtotrader:
This PR should close #335

femtotrader:
For PEP8 you should write

    .replace(',', '')

instead of

    .replace(',','')

jreback:
thanks.

PR #336, opened 2017-05-18
[WIP] Update conda build recipe for v0.4.0

ralexx:
The `meta.yaml` in the v0.4.0 release has not been updated.

This recipe builds on my macOS system but only if the `pandas_datareader.io.test` import is omitted. That module is (still) in the source and has an `__init__.py` so I don't understand why it is raising.

```python
import: 'pandas_datareader.io.tests'
Traceback (most recent call last):
  File "/usr/local/share/anaconda/conda-bld/pandas-datareader_1495128936984/test_tmp/run_test.py", line 11, in <module>
    import pandas_datareader.io.tests
ModuleNotFoundError: No module named 'pandas_datareader.io.tests'
TESTS FAILED: pandas-datareader-0.4.0-py36_0.tar.bz2
```

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -1,11 +1,11 @@
 package:
   name: pandas-datareader
-  version: "0.2.0"
+  version: "0.4.0"
 
 source:
-  fn: pandas-datareader-0.2.0.tar.gz
-  url: https://pypi.python.org/packages/source/p/pandas-datareader/pandas-datareader-0.2.0.tar.gz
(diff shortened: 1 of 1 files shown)


femtotrader:
Maybe some details should be add to `release-procedure.md`

Pinging @jreback @davidastephens

jreback:
this recipe is not actually used, going to remove it. instead it is built (automatically) via https://github.com/conda-forge/pandas-datareader-feedstock/blob/master/recipe/meta.yaml

PR #346, opened 2017-06-12
added QuandlReader (plus tests and doc)

prechelt:
A reasonably nice Quandl adapter (more needed now that the Yahoo Finance daily data API has gone down).

Because of the Yahoo breakdown, the tests currently do not run.
Also, I am on Windows, where the Google tests will be skipped.
But I have probably not broken anything; I've limited my changes to existing parts to the minimum.

Note: One should refactor the url/_get_params construction for BaseReader: The format of the URL belongs under the complete control of the implementing class. For Quandl this is needed because the DB name is part of the base URL, not the parameters. My current construction is a bit of a kludge in this respect.


prechelt:
Regarding Yahoo Finance:
This thread let me conclude it is not likely to come back:
https://forums.yahoo.net/t5/Yahoo-Finance-help/Is-Yahoo-Finance-API-broken/td-p/250503

For the time being, I suggest to deactivate the tests, put a remark into the docs, but leave the code in place (and remove it in a year or so).

jreback:
can you rebase

jreback:
superseded by #361

PR #347, opened 2017-06-13
fix google date formatting

richyliu:
The Google finance api formats its dates with year-month-date, not month day, year.

### DIFF
--- a/pandas_datareader/google/daily.py
+++ b/pandas_datareader/google/daily.py
@@ -36,8 +36,8 @@ class GoogleDailyReader(_DailyBaseReader):
     def _get_params(self, symbol):
         params = {
             'q': symbol,
-            'startdate': self.start.strftime('%b %d, %Y'),
-            'enddate': self.end.strftime('%b %d, %Y'),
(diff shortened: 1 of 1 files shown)


davidastephens:
Is there a test that fails now and passes with this change?

bsolomon1124:
These needs to be addressed.  The url format has very clearly changed: 

> https://finance.google.com/finance/historical?q=SPY&startdate=2017/01/01&enddate=2017/05/22&output=csv

is now valid.

bsolomon1124:
It seems like the issue may not be with the date formatting.  Both of these work:

**Old format:**
http://finance.google.com/finance/historical?q=SPY&startdate=Jan+01%2C+2017&enddate=Dec+01%2C+2017&output=csv

**New format (proposed here):**

> http://finance.google.com/finance/historical?q=googl&startdate=2010%2F01%2F01&enddate=2017%2F12%2F01&output=csv

davidastephens:
Similar to #425, but both links work per @bsolomon1124.  Closing.

PR #357, opened 2017-07-03
fix 'null' in lower case as a  missing value in the new Yahoo API, make compatible with the older pandas versions

### LINKED ISSUE #342: Exception downloading yahoo historical data with  adjust_price=True on  fix-yahoo branch
Exception downloading yahoo historical data with  adjust_price=True on  fix-yahoo branch

This can be recreated for symbol = 'SRCE' and date range should include Jun 29, 2016

yahooData = pdr.data.get_data_yahoo(symbol, self.start, self.end, adjust_price=True)
File "/home/oleg/projects/openSource/python/pandas-datareader/yahoofix/pandas-datareader/pandas_datareader/data.py", line 40, in get_data_yahoo
return YahooDailyReader(*args, **kwargs).read()
File "/home/oleg/projects/openSource/python/pandas-datareader/yahoofix/pandas-datareader/pandas_datareader/yahoo/daily.py", line 117, in read
df = _adjust_prices(df)
File "/home/oleg/projects/openSource/python/pandas-datareader/yahoofix/pandas-datareader/pandas_datareader/yahoo/daily.py", line 170, in _adjust_prices
adj_ratio = hist_data['Adj Close'] / hist_data['Close']
File "/home/oleg/programs/python/anaconda2/envs/yahoofix/lib/python2.7/site-packages/pandas/core/ops.py", line 721, in wrapper
result = wrap_results(safe_na_op(lvalues, rvalues))
File "/home/oleg/programs/python/anaconda2/envs/yahoofix/lib/python2.7/site-packages/pandas/core/ops.py", line 692, in safe_na_op
lambda x: op(x, rvalues))
File "pandas/_libs/algos_common_helper.pxi", line 1212, in pandas._libs.algos.arrmap_object (pandas/_libs/algos.c:31954)
File "/home/oleg/programs/python/anaconda2/envs/yahoofix/lib/python2.7/site-packages/pandas/core/ops.py", line 692, in <lambda>
lambda x: op(x, rvalues))
TypeError: unsupported operand type(s) for /: 'str' and 'str'

Environment:
followed the instruction found on one of the forums probably on stackoverflow:
git clone https://github.com/rgkimball/pandas-datareader
$ cd pandas-datareader
$ git checkout fix-yahoo
$ pip install -e .

It turned out that on yahoo web site the data displayed:
Jun 30, 2016 31.73 32.44 31.65 32.39 32.39 28,600
Jun 29, 2016 0.00 0.00 0.00 0.00 0.00 -
Jun 28, 2016 31.10 31.16 30.63 31.05 31.05 49,800

but when you click on the download button you will get:
2016-06-28 31.1 31.16 30.629999 31.049999 31.049999 49800
2016-06-29 null null null null null null
2016-06-30 31.73 32.439999 31.65 32.389999 32.389999 28600

string 'null' instead of 0.00 or '-'
The same lower case 'null' values will be if you use  adjust_price=False in pdr.data.get_data_yahoo

The solution is to add 'null' in lower case to na_values argument to read_csv in function _read_lines in base.py
na_values=('-', 'null')

I have made this change that is my repository.

OlegShteynbuk:
closes #342
 
add 'null' in lower case to na_values argument to read_csv

### DIFF
--- a/pandas_datareader/tests/yahoo/test_yahoo.py
+++ b/pandas_datareader/tests/yahoo/test_yahoo.py
@@ -133,8 +133,8 @@ class TestYahoo(object):
 
     @pytest.mark.parametrize('adj_pr', [True, False])
     def test_get_data_null_as_missing_data(self, adj_pr):
-        # workaround as skip_on_exception decorator changes signature
-        # TODO: should it be changed to a signature-preserving decorator?
+        # TODO: We can't decorate the test function along with
+        # parametrization because it errors.  Investigate this later.
         @skip_on_exception(RemoteDataError)
         def null_as_missing_data(adj_price):
             result = web.get_data_yahoo('SRCE', '20160626', '20160705',
--- a/pandas_datareader/tests/yahoo/test_yahoo.py
+++ b/pandas_datareader/tests/yahoo/test_yahoo.py
@@ -134,7 +134,7 @@ class TestYahoo(object):
     @pytest.mark.parametrize('adj_pr', [True, False])
     def test_get_data_null_as_missing_data(self, adj_pr):
         # workaround as skip_on_exception decorator changes signature
-        # should it be changed to a signature-preserving decorator?
+        # TODO: should it be changed to a signature-preserving decorator?
         @skip_on_exception(RemoteDataError)
         def null_as_missing_data(adj_price):
             result = web.get_data_yahoo('SRCE', '20160626', '20160705',
--- a/pandas_datareader/tests/yahoo/test_yahoo.py
+++ b/pandas_datareader/tests/yahoo/test_yahoo.py
@@ -132,6 +132,11 @@ class TestYahoo(object):
         web.get_data_yahoo(sl, '2012')
 
     @skip_on_exception(RemoteDataError)
+    def test_get_data_null_as_missing_data(self):
+        # just test that we succeed
+        web.get_data_yahoo('SRCE', '20160626', '20160705', adjust_price=True)
+
+    @skip_on_exception(RemoteDataError)
     def test_get_data_multiple_symbols_two_dates(self):
         pan = web.get_data_yahoo(['GE', 'MSFT', 'INTC'], 'JAN-01-12',
                                  'JAN-31-12')
--- a/pandas_datareader/tests/yahoo/test_yahoo.py
+++ b/pandas_datareader/tests/yahoo/test_yahoo.py
@@ -131,17 +131,18 @@ class TestYahoo(object):
         sl = ['AAPL', 'AMZN', 'GOOG']
         web.get_data_yahoo(sl, '2012')
 
-    @skip_on_exception(RemoteDataError)
-    def test_get_data_null_as_missing_data_no_adjust(self):
-        result = web.get_data_yahoo('SRCE', '20160626', '20160705',
-                                    adjust_price=False)
-        # sanity checking
-        assert result.dtypes.all() == np.floating
+    @pytest.mark.parametrize('adj_pr', [True, False])
+    def test_get_data_null_as_missing_data(self, adj_pr):
+        #workaround as skip_on_exception decorator changes signature
+        # should it be changed to a signature-preserving decorator?
+        @skip_on_exception(RemoteDataError)
+        def null_as_missing_data(adj_price):
+            result = web.get_data_yahoo('SRCE', '20160626', '20160705',
+                                        adjust_price=adj_pr)
+            # sanity checking
+            assert result.dtypes.all() == np.floating
 
-    @skip_on_exception(RemoteDataError)
-    def test_get_data_null_as_missing_data_adjust(self):
-        # just test that we succeed
-        web.get_data_yahoo('SRCE', '20160626', '20160705', adjust_price=True)
+        null_as_missing_data(adj_pr)
 
     @skip_on_exception(RemoteDataError)
     def test_get_data_multiple_symbols_two_dates(self):
--- a/pandas_datareader/tests/yahoo/test_yahoo.py
+++ b/pandas_datareader/tests/yahoo/test_yahoo.py
@@ -133,7 +133,6 @@ class TestYahoo(object):
 
     @skip_on_exception(RemoteDataError)
     def test_get_data_null_as_missing_data_no_adjust(self):
-        # just test that we succeed
         result = web.get_data_yahoo('SRCE', '20160626', '20160705',
                                     adjust_price=False)
         # sanity checking
--- a/pandas_datareader/tests/yahoo/test_yahoo.py
+++ b/pandas_datareader/tests/yahoo/test_yahoo.py
@@ -131,6 +131,7 @@ class TestYahoo(object):
         sl = ['AAPL', 'AMZN', 'GOOG']
         web.get_data_yahoo(sl, '2012')
 
+    @skip_on_exception(RemoteDataError)
     def test_get_data_null_as_missing_data_no_adjust(self):
         # just test that we succeed
         result = web.get_data_yahoo('SRCE', '20160626', '20160705',
--- a/pandas_datareader/tests/yahoo/test_yahoo.py
+++ b/pandas_datareader/tests/yahoo/test_yahoo.py
@@ -133,7 +133,8 @@ class TestYahoo(object):
 
     def test_get_data_null_as_missing_data_no_adjust(self):
         # just test that we succeed
-        result = web.get_data_yahoo('SRCE', '20160626', '20160705', adjust_price=False)
+        result = web.get_data_yahoo('SRCE', '20160626', '20160705',
+                                    adjust_price=False)
         # sanity checking
         assert result.dtypes.all() == np.floating
 
--- a/pandas_datareader/tests/yahoo/test_yahoo.py
+++ b/pandas_datareader/tests/yahoo/test_yahoo.py
@@ -133,7 +133,7 @@ class TestYahoo(object):
 
     @pytest.mark.parametrize('adj_pr', [True, False])
     def test_get_data_null_as_missing_data(self, adj_pr):
-        #workaround as skip_on_exception decorator changes signature
+        # workaround as skip_on_exception decorator changes signature
         # should it be changed to a signature-preserving decorator?
         @skip_on_exception(RemoteDataError)
         def null_as_missing_data(adj_price):
--- a/pandas_datareader/tests/yahoo/test_yahoo.py
+++ b/pandas_datareader/tests/yahoo/test_yahoo.py
@@ -131,8 +131,14 @@ class TestYahoo(object):
         sl = ['AAPL', 'AMZN', 'GOOG']
         web.get_data_yahoo(sl, '2012')
 
+    def test_get_data_null_as_missing_data_no_adjust(self):
+        # just test that we succeed
+        result = web.get_data_yahoo('SRCE', '20160626', '20160705', adjust_price=False)
+        # sanity checking
+        assert result.dtypes.all() == np.floating
+
     @skip_on_exception(RemoteDataError)
-    def test_get_data_null_as_missing_data(self):
+    def test_get_data_null_as_missing_data_adjust(self):
         # just test that we succeed
         web.get_data_yahoo('SRCE', '20160626', '20160705', adjust_price=True)
 
--- a/docs/source/whatsnew/v0.5.0.txt
+++ b/docs/source/whatsnew/v0.5.0.txt
@@ -23,3 +23,4 @@ Bug Fixes
 - Test suite fixes for test_get_options_data (:issue:`352`)
 - Test suite fixes for test_wdi_download (:issue:`350`)
 - avoid monkey patching requests.Session (:issue:`301`)
+- :func:`get_data_yahoo` now treats ``'null'`` strings as missing values  (:issue:`342`)
--- a/pandas_datareader/base.py
+++ b/pandas_datareader/base.py
@@ -143,7 +143,7 @@ class _BaseReader(object):
         raise NotImplementedError("Subclass has not implemented method.")
 
     def _read_lines(self, out):
-        rs = read_csv(out, index_col=0, parse_dates=True, na_values='-')[::-1]
+        rs = read_csv(out, index_col=0, parse_dates=True, na_values=('-', 'null'))[::-1]
         # Yahoo! Finance sometimes does this awesome thing where they
         # return 2 rows for the most recent business day
         if len(rs) > 2 and rs.index[-1] == rs.index[-2]:  # pragma: no cover
--- a/pandas_datareader/base.py
+++ b/pandas_datareader/base.py
@@ -143,7 +143,8 @@ class _BaseReader(object):
         raise NotImplementedError("Subclass has not implemented method.")
 
     def _read_lines(self, out):
-        rs = read_csv(out, index_col=0, parse_dates=True, na_values=('-', 'null'))[::-1]
+        rs = read_csv(out, index_col=0, parse_dates=True,
+                      na_values=('-','null'))[::-1]
         # Yahoo! Finance sometimes does this awesome thing where they
         # return 2 rows for the most recent business day
         if len(rs) > 2 and rs.index[-1] == rs.index[-2]:  # pragma: no cover
--- a/pandas_datareader/base.py
+++ b/pandas_datareader/base.py
@@ -144,7 +144,7 @@ class _BaseReader(object):
 
     def _read_lines(self, out):
         rs = read_csv(out, index_col=0, parse_dates=True,
-                      na_values=('-','null'))[::-1]
+                      na_values=('-', 'null'))[::-1]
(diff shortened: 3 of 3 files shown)


jreback:
can u add the above test so that we can see that this works

OlegShteynbuk:
add 'null' in lower case to na_values arguments

OlegShteynbuk:
added test

pandas_datareader/tests/yahoo/test_yahoo.py
```diff
@@ -132,6 +132,11 @@ def test_get_data_multiple_symbols(self):
         web.get_data_yahoo(sl, '2012')
 
     @skip_on_exception(RemoteDataError)
+    def test_get_data_null_as_missing_data(self):
+        # just test that we succeed
```
jreback: add both cases and assert the dtypes of the result

OlegShteynbuk:
a note in the whatsnew for 0.5 added

pandas_datareader/tests/yahoo/test_yahoo.py
OlegShteynbuk: I was thinking about the second test but realized that in the future pandas versions (0.21.0 and up)  `read_csv` will return 'np.nan' and all previous versions will return 'null'. We can check the pandas version but it is not yet released version not sure if there is a chance of the version number change.  Also, the test will depend on Yahoo data handling and they recently started to use 'null' as a missing value and can change it in the future. We can mock Yahoo data or pandas  `read_csv` function and in the latter case, there will be no need to check pandas version.

OlegShteynbuk: added

pandas_datareader/tests/yahoo/test_yahoo.py
```diff
@@ -132,6 +132,18 @@ def test_get_data_multiple_symbols(self):
         web.get_data_yahoo(sl, '2012')
 
     @skip_on_exception(RemoteDataError)
+    def test_get_data_null_as_missing_data_no_adjust(self):
```
jreback: parametrize this with adjust_prices=True & False (and remove the test below)

jreback:
FYI we are testing on older pandas versions and such so no special handling, everything should work. null should be fine in newer (as its default) and older (as you added it).

OlegShteynbuk:
Added parametrize. In order to use parametrize created an inner function(added a level of indirection) because skip_on_exception decorator changes the signature. The other solution would be to make skip_on_exception decorator a signature-preserving decorator. But this is outside of the scope of this PR as it will affect other tests; my first impression is that it will affect other tests for the better as will allow using skip_on_exception decorator with parametrize, it could  be a little bit slower but the difference is rather small and for testing it should not be an issue. I can go ahead and create a new issue.

pandas_datareader/tests/yahoo/test_yahoo.py
```diff
@@ -133,7 +133,7 @@ def test_get_data_multiple_symbols(self):
 
     @pytest.mark.parametrize('adj_pr', [True, False])
     def test_get_data_null_as_missing_data(self, adj_pr):
-        #workaround as skip_on_exception decorator changes signature
```
bashtage: Does using a fixture fix this?

```
@pytest.fixture(params=[True,False], scope='module')
def adj_pr(request)
    return request.param

@skip_on_exception(RemoteDataError)
def test_get_data_null_as_missing_data(self, adj_pr):
    ...
```

OlegShteynbuk: @bashtage    thanks for the code sample interesting usage of fixtures, wasn't able to run it successfully,  but I am no pytest expert just a casual user; also this approach has an additional level of indirection(a new fixture) and probably some advanced pytest knowledge but a signature-preserving decorator hopefully should work just like any other decorator, just put one line in front of a function

pandas_datareader/tests/yahoo/test_yahoo.py
```diff
@@ -131,6 +131,19 @@ def test_get_data_multiple_symbols(self):
         sl = ['AAPL', 'AMZN', 'GOOG']
         web.get_data_yahoo(sl, '2012')
 
+    @pytest.mark.parametrize('adj_pr', [True, False])
+    def test_get_data_null_as_missing_data(self, adj_pr):
+        # workaround as skip_on_exception decorator changes signature
+        # should it be changed to a signature-preserving decorator?
```
gfyoung: Decorators generally are difficult to make signature-preserving AFAIK.  Does `pytest` and this decorator not work well together?  Did you try using `skip_on_exception` above the `pytest.mark` one?

OlegShteynbuk: Yes, it's true that signature-preserving decorators are difficult to make but there are at least two packages that I was planning on trying.

I have tried all combinations of `skip_on_exception'`and `the pytest.mark` above and below.

gfyoung: What happens when you try either? (i.e. what stacktrace do you get?)

pandas_datareader/tests/yahoo/test_yahoo.py
gfyoung: I think @bashtage gave you everything there, so there is nothing else to do but copy and paste.  What happens when you do this?

pandas_datareader/tests/yahoo/test_yahoo.py
OlegShteynbuk: bellow is the error message

E           TypeError: test_null_as_missing_data() takes exactly 2 arguments (1 given)

OlegShteynbuk: maybe it will work with global functions not when they are inside the class, not sure

gfyoung: Hmm...odd that this occurs in both cases.  To confirm, do you get the same error regardless of the order?

OlegShteynbuk: yes

gfyoung: Hmm...in that case, change the comment and just add a "TODO" saying that we should investigate this signature incompatibility later.

OlegShteynbuk: btw, not sure if providing a fixture is a better solution than an inner function, at least to me the inner function solution looks simpler.

gfyoung: A fixture is more idiomatic, but then again, you said had issues with that, yes?

OlegShteynbuk: yes, i do; will try some more experiments some time later

pandas_datareader/tests/yahoo/test_yahoo.py
```diff
@@ -134,7 +134,7 @@ def test_get_data_multiple_symbols(self):
     @pytest.mark.parametrize('adj_pr', [True, False])
     def test_get_data_null_as_missing_data(self, adj_pr):
         # workaround as skip_on_exception decorator changes signature
-        # should it be changed to a signature-preserving decorator?
+        # TODO: should it be changed to a signature-preserving decorator?
```
gfyoung: Given that signature preservation is difficult, I think a better message would be this:
~~~python
def test_..
  # TODO: We can't decorate the test function on top of parametrization because it errors.  Investigate this later.
~~~

OlegShteynbuk: i have tried both ways on top and below

gfyoung:
~~~python
def test_..
  # TODO: We can't decorate the test function along with
  # parametrization because it errors.  Investigate this later.
~~~

OlegShteynbuk: done, thanks; it's definitely better comments

bashtage:
FWIW, this works as expected

Outside the class

```
@pytest.fixture(params=[True, False])
def adj_pr(request):
    return request.param
```

And the test function -- obviously one would remove the intentional `        raise RemoteDataError('This is a test')`

```
    @skip_on_exception(RemoteDataError)
    def test_get_data_null_as_missing_data(self, adj_pr):
        result = web.get_data_yahoo('SRCE', '20160626', '20160705',
                                    adjust_price=adj_pr)
        # sanity checking
        assert result.dtypes.all() == np.floating
        raise RemoteDataError('This is a test')
```

pandas_datareader/tests/yahoo/test_yahoo.py
jreback: yeah have to do something like this: https://stackoverflow.com/questions/12197263/pytest-parameterize-decorator-not-working-with-kwargs

@gfyoung if you want to fix up the skip_on_exception would be great (create an issue to start if you would)

gfyoung: @jreback : Certainly.  I can't attend to it now, but I can soon (I'll also want to wait until this gets merged before doing so).

jreback:
@OlegShteynbuk if you want to change to use a fixture as @bashtage suggest would be great. ping on green.

pandas_datareader/tests/yahoo/test_yahoo.py
OlegShteynbuk: sure, I have look at the decorator package not sure if can use it in our case, but there is another one that i can use too, will create an issue

jreback: yep no hurry at all.

gfyoung: > I have looked at the decorator package not sure how to use it in our case,

IMO, using another package seems a bit overkill.  I imagine that the fix for this issue shouldn't need to require another package just for this one decorator.

OlegShteynbuk: sorry, I didn't make it clear but it meant instead of a decorator package, one or the other, not an extra one in addition to the decorator package

gfyoung: I'm not sure that actually clarified what you were saying...but no worries, so long as the PR gets emerged eventually :smile:

OlegShteynbuk:
I have tried bashtage suggestion before with the fixture outside the class and inside the class, not sure what was the problem, maybe my environment is different, I have python  2.7.13  and pytest 3.0.7 on LInux, not sure what was his testing environment, maybe python 3; need to take a break... 

It could be an interesting psychological experiment to find out if a preference for the solution depends on one's background; for me coming from Java and C++ encapsulation is a big deal and having a nested function makes it all nicely encapsulated in one place instead of having two functions in two different places; but it's always a good idea to look at the same problem from a different point of view, I already learned a lot about pytest from this discussion, thanks.

gfyoung:
I think in this case I would go with what @bashtage implemented.  Test fixtures are meant to be used across tests, so it makes sense to share this resource and not encapsulate.  We have plenty of instances of encapsulation (functions within functions) in the codebase.

Also, I'm a little surprised you went for inner functions given your Java and C++ background, since IINM, neither supports explicit declarations of inner functions.  Sure, you have lambdas, but they are anonymous functions instead of the explicit one that you defined currently.

OlegShteynbuk:
I have repeated the @bashtage test with pretty much the same results, doesn't run in my environment, below is the output and i have run it beforein different combinations, there are actually two problems as it says `collected 21 items` but it should be 22, and if you look at the output  there is only one record for `test_get_data_null_as_missing_data` instead of two for True and False
pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_get_data_null_as_missing_data <- pandas_datareader/_testing.py FAILED

below is the output

========================================================================


platform linux2 -- Python 2.7.13, pytest-3.0.7, py-1.4.31, pluggy-0.4.0 -- /home/oleg/programs/python/anaconda2/envs/pandas-datareader_dev/bin/python
cachedir: .cache
rootdir: /home/oleg/projects/openSource/python/pandas-datareader/pandas-datareader, inifile:
plugins: xdist-1.16.0
collected 21 items 

pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_yahoo <- pandas_datareader/_testing.py PASSED
pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_get_data_multiple_symbols <- pandas_datareader/_testing.py PASSED
pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_get_data_adjust_price <- pandas_datareader/_testing.py PASSED
pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_get_data_multiple_symbols_two_dates <- pandas_datareader/_testing.py PASSED
pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_get_data_single_symbol <- pandas_datareader/_testing.py PASSED
pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_yahoo_DataReader_multi <- pandas_datareader/_testing.py PASSED
pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_yahoo_reader_class <- pandas_datareader/_testing.py PASSED
pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_get_data_null_as_missing_data <- pandas_datareader/_testing.py FAILED
pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_yahoo_fails PASSED
pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_get_quote_series PASSED
pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_get_quote_string PASSED
pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_get_quote_stringlist PASSED
pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_get_quote_comma_name PASSED
pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_get_components_dow_jones SKIPPED
pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_get_components_dax SKIPPED
pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_get_components_nasdaq_100 SKIPPED
pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_get_data_interval xfail
pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_get_date_ret_index XPASS
pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_get_data_yahoo_actions xfail
pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_get_data_yahoo_actions_invalid_symbol PASSED
pandas_datareader/tests/yahoo/test_yahoo.py::TestYahoo::test_yahoo_DataReader xfail

==================================================================================================== FAILURES ====================================================================================================
__________________________________________________________________________________ TestYahoo.test_get_data_null_as_missing_data __________________________________________________________________________________

args = (<pandas_datareader.tests.yahoo.test_yahoo.TestYahoo object at 0x7f03d1549890>,), kwargs = {}

    @wraps(f)
    def wrapper(*args, **kwargs):
        try:
>           f(*args, **kwargs)
E           TypeError: test_get_data_null_as_missing_data() takes exactly 2 arguments (1 given)

pandas_datareader/_testing.py:26: TypeError
===================================================================== 1 failed, 13 passed, 3 skipped, 3 xfailed, 1 xpassed in 41.76 seconds ======================================================================

OlegShteynbuk:
> Also, I'm a little surprised you went for inner functions given your Java and C++ background, since IINM, neither supports explicit declarations of inner functions. Sure, you have lambdas, but they are anonymous functions instead of the explicit one that you defined currently.
=======================

there are also local and anonymous classes in Java that can be used to define a member function.
Was thinking more about the programming style that I was using most of the time in Java/C++, and also looked at this task as a standalone task not intended for reuse.

OlegShteynbuk:
on a different note, probably a newbie github question, what does `ping on green` means, don't see any green buttons available for me to push on.

gfyoung:
> there are also local and anonymous classes in Java that can be used to define a member function

But they're anonymous nonetheless :smile:

What @jreback means by pinging on green is this: when all of the tests pass (as they have now), just write another comment like this:

@jreback : Everything is green and ready to go!

OlegShteynbuk:
@jreback : Everything is green and ready to go!

bashtage:
@OlegShteynbuk is right -- Python 2.7 doesn't correctly preserve the function signature when decorated, and so pytest doesn't realize that it needs pass the fixture in.  This results in the wrong number of arguments being passed, and an error occurs.  

Here it fails (2.7)

https://travis-ci.org/bashtage/pandas-datareader/builds/250677402

While it works fine on all Python 3.x builds. Yet another reason for 2.7 pass to oblivion.

jreback:
thanks!

pls open an issue to change the decorator as well.

OlegShteynbuk:
issue https://github.com/pydata/pandas-datareader/issues/363

Issue #356, opened 2017-07-02
followup on Yahoo DataReader issues


jreback:
in #355 I xfailed a couple of tests as they were not pulling data. These should be investigated before the release.

cc @rgkimball

jrovegno:
When I put
`pdr.get_data_yahoo('SPY')`
I get this error
`ConnectionError: HTTPConnectionPool(host='ichart.finance.yahoo.com', port=80): Max retries exceeded with url: /table.csv?s=SPY&a=0&b=1&c=2010&d=6&e=4&f=2017&g=d&ignore=.csv (Caused by NewConnectionError('<requests.packages.urllib3.connection.HTTPConnection object at 0x0000000009DE1198>: Failed to establish a new connection: [Errno 11004] getaddrinfo failed',))`

Similar [problem in  the docs](http://pandas-datareader.readthedocs.io/en/latest/remote_data.html#remote-data-yahoo).

gliptak:
@jreback Do you plan to  publish 0.5.0 soon? Thanks

https://pypi.python.org/pypi/pandas-datareader

jreback:
yes if this can be resolved soon

gliptak:
I cannot reproduce this above with current ```master``` (installed as ```pip install git+git://github.com/pydata/pandas-datareader.git```):

```
➜ pip list | grep pandas-datareader
pandas-datareader (0.5.0)
➜ jupyter-console                  
Jupyter console 5.1.0

Python 2.7.12+ (default, Sep 17 2016, 12:08:02) 
Type "copyright", "credits" or "license" for more information.

IPython 5.4.1 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.



In [1]: import pandas_datareader as pdr

In [2]: pdr.get_data_yahoo('SPY').tail()
Out[2]: 
                  Open        High         Low       Close   Adj Close  \
Date                                                                     
2017-06-27  243.039993  243.380005  241.309998  241.330002  241.330002   
2017-06-28  242.500000  243.720001  242.229996  243.490005  243.490005   
2017-06-29  243.660004  243.720001  239.960007  241.350006  241.350006   
2017-06-30  242.279999  242.710007  241.580002  241.800003  241.800003   
2017-07-03  242.880005  243.380005  242.210007  242.210007  242.210007   

               Volume  
Date                   
[...]

PR #365, opened 2017-07-08
Correct Yahoo period end timestamp calculation

gliptak:

xref https://github.com/pydata/pandas-datareader/issues/356

### DIFF
--- a/pandas_datareader/yahoo/daily.py
+++ b/pandas_datareader/yahoo/daily.py
@@ -97,7 +97,8 @@ class YahooDailyReader(_DailyBaseReader):
 
     def _get_params(self, symbol):
         unix_start = int(time.mktime(self.start.timetuple()))
-        unix_end = int(time.mktime(self.end.timetuple()))
+        day_end = self.end.replace(hour=23, minute=59, second=59)
+        unix_end = int(time.mktime(day_end.timetuple()))
 
         params = {
             'period1': unix_start,
--- a/pandas_datareader/tests/yahoo/test_yahoo.py
+++ b/pandas_datareader/tests/yahoo/test_yahoo.py
@@ -104,17 +104,17 @@ class TestYahoo(object):
         assert 'Adj Close' not in goog_adj.columns
         assert (goog['Open'] * goog_adj['Adj_Ratio']).equals(goog_adj['Open'])
 
-    @pytest.mark.xfail(reason="failing after #355")
+    @skip_on_exception(RemoteDataError)
     def test_get_data_interval(self):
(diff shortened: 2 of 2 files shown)


pandas_datareader/yahoo/daily.py
```diff
@@ -97,7 +98,9 @@ def yurl(symbol):
 
     def _get_params(self, symbol):
         unix_start = int(time.mktime(self.start.timetuple()))
-        unix_end = int(time.mktime(self.end.timetuple()))
+        day_end = self.end + datetime.timedelta(days=1) \
```
bashtage: Is it safe to assume self.end is a whole day only with time 0:00:00?

gliptak: It is a datetime created from year/month/day (as discussed in https://github.com/pydata/pandas-datareader/issues/356 there still might be some timezone issues ...)
The yahoo API becoming very sensitive ...

jreback: does this need any additional tests?

gliptak: @jreback I'm proposing this change, because I noted that existing Yahoo tests behaved differently locally vs on Travis. @bashtage also seem to have observed same. I see existing tests covering this change (although maybe somebody could reconfirm boundaries?)

bashtage: This approach is not correct.

```
r = pdr.yahoo.daily.YahooDailyReader(['IBM'],'1/1/1960','1/1/2017 12:30:00')
r.end
Out[6]: Timestamp('2017-01-01 12:30:00')
```

Since it accepts times as well as dates you will need to create a new timestamp using the year, day and month and setting the hour to 23, minute to 59, and second to 59.

gliptak: updated

gliptak:
@jreback Green after updates

pandas_datareader/yahoo/daily.py
jreback: I would add a test like this, to assert the end-points of the received data are correct (should fail before and pass after this PR)

```
r = pdr.yahoo.daily.YahooDailyReader(['IBM'],'1/1/1960','1/1/2017 12:30:00')
r.end
```

gliptak: @jreback As discussed above, the failure was intermittent, sometimes not returning the data for 12/31. I'M m not sure what value a test on ```r.end``` adds ... Thanks

jreback:
thanks @gliptak

Issue #396, opened 2017-09-16
Incorrect split calculation in yahoo-actions


tibkiss:
When yahoo-actions is used to fetch split information then the resulting value is either 0 or raises ZeroDivisionError. 

YahooActionReader tries to eval the returned string then take the reciprocal of the evaluated statement: 
https://github.com/pydata/pandas-datareader/blob/331e3895c51326e87e69325e1a14f1e350cb832e/pandas_datareader/yahoo/actions.py#L35

If the split results reduced number of stocks then the eval'd value will be correct, but the reciprocal calculation will result 0 as integers are used by the operation:
```
In [24]: 1/eval('2/1')
Out[24]: 0
```
This issue could be fixed by specifying 1.0 in the numerator.

If the split results increased number of stocks then the `eval()` will yield a zero, which is used in the denominator causing ZeroDivisionError:
```
In [26]: 1/eval('1/2')
---------------------------------------------------------------------------
ZeroDivisionError                         Traceback (most recent call last)
<ipython-input-26-dc71badc23e7> in <module>()
----> 1 1/eval('1/2')

ZeroDivisionError: integer division or modulo by zero
```
One possible fix could be to parse the string, convert the numerator and denominator to float then calculate the proper split value.

Env: pandas-datareader==0.5.0, Python 2.7

bashtage:
Yahoo Actions has been deprecated.

PR #397, opened 2017-09-16
Fix Split calculation in YahooActionReader

tibkiss:
See https://github.com/pydata/pandas-datareader/issues/396 for details.

### DIFF
--- a/pandas_datareader/yahoo/actions.py
+++ b/pandas_datareader/yahoo/actions.py
@@ -1,3 +1,5 @@
+from __future__ import division
+
 from pandas import (concat, DataFrame)
 from pandas_datareader.yahoo.daily import YahooDailyReader
 


bashtage:
Should have a comment with the GH issue number that identified it (# GH #396)

bashtage:
Any reason not to just import 

from __future__ import division

?

tibkiss:
@bashtage : Thanks for the prompt review, update the PR accordingly.

tibkiss:
The errors seems unrelated to me. Could someone please confirm too?

tibkiss:
@gliptak : PTAL

tibkiss:
I've encountered a ZeroDivisionError for a given security in the same code path. Shall I include handling such errors in this PR or rather open a new one?

bashtage:
Fixing in this PR would be good.

tibkiss:
will open a new pr.

Issue #398, opened 2017-09-16
yahoo-actions downloader fails when there were no splits in the given timeframe


tibkiss:
YahooActionReader fails if there were no splits for the instrument on the given time-frame:

```
In [5]: DataReader('AAPL', 'yahoo-actions', datetime.datetime(2009, 1, 1), datetime.datetime(2010, 1, 1))
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-5-34d77b66d048> in <module>()
----> 1 a = DataReader('AAPL', 'yahoo-actions', start, end)

/Users/tiborkiss/devel/workspace/stocks/venv-zipline-live-jaccos/lib/python2.7/site-packages/pandas_datareader/data.pyc in DataReader(name, data_source, start, end, retry_count, pause, session, access_key)
    124         return YahooActionReader(symbols=name, start=start, end=end,
    125                                  retry_count=retry_count, pause=pause,
--> 126                                  session=session).read()
    127     elif data_source == "yahoo-dividends":
    128         return YahooDivReader(symbols=name, start=start, end=end,

/Users/tiborkiss/devel/workspace/stocks/venv-zipline-live-jaccos/lib/python2.7/site-packages/pandas_datareader/yahoo/actions.pyc in read(self)
     33             # Converts fractional form splits (i.e. "2/1") into conversion
     34             # ratios, then take the reciprocal
---> 35             splits['value'] = splits.apply(lambda x: 1/eval(x['value']), axis=1)  # noqa
     36
     37         output = concat([dividends, splits]).sort_index(ascending=False)

/Users/tiborkiss/devel/workspace/stocks/venv-zipline-live-jaccos/lib/python2.7/site-packages/pandas/core/frame.pyc in __setitem__(self, key, value)
   2329         else:
   2330             # set column
-> 2331             self._set_item(key, value)
   2332
   2333     def _setitem_slice(self, key, value):

/Users/tiborkiss/devel/workspace/stocks/venv-zipline-live-jaccos/lib/python2.7/site-packages/pandas/core/frame.pyc in _set_item(self, key, value)
   2394         """
   2395
-> 2396         self._ensure_valid_index(value)
   2397         value = self._sanitize_column(key, value)
   2398         NDFrame._set_item(self, key, value)

[...]

PR #399, opened 2017-09-16
Support download of yahoo actions containing no split events (#398)

tibkiss:


tibkiss:
@gliptak : PTAL

tibkiss:
will open a new pr

PR #403, opened 2017-09-27
WIP: Fix tests on the master branch

soon:
The current master branch does not pass all tests

This PR is going to fix them

PR #402, opened 2017-09-21
WIP: Issue #394: Fixed google url

soon:
Issue #394

WIP until tests on the master branch will be fixed: https://github.com/pydata/pandas-datareader/pull/403

Issue #395, opened 2017-09-16
Pandas data reader for google  end dates are not working 


nithints:
import pandas_datareader.data as web
 import datetime
 start = datetime.datetime(2010, 1, 1)

 end = datetime.datetime(2015, 5, 9)

 web.DataReader('AAPL', 'google', start, end)

getting an year data irrespective of start and end dates
![image](https://user-images.githubusercontent.com/6574109/30509594-270f6644-9ad2-11e7-9999-205e5aed5811.png)

rsvp:
- [ ] **Duplicate** issue of #391 and #394

bkcollection:
@rsvp  Is the google historical data is just merely a change of link from www.google.com to finance.google.com ?
or it has a more complicated issues to deal with before merge into pandas-datareader? If it is stopped will cause a big issues for non US user as we do not have any alternatives which is need to cost a lot of money like Bloomberg/reuters

rsvp:
@bkcollection I have covered both *Google Finance URL Change* and *Alternative Data Sources* at https://github.com/rsvp/fecon235/issues/7 in detail. The situation is still in flux.

1rocketdude:
thank you for the thorough description of the problem with Google and
Yahoo. I've had good luck with Quandl, but it is not infallible. Some
symbols are not available. Have not tried Tingo and Alpha Advantage, so I
have no experience.

Looks like time to ditch Google and Yahoo. Frequent interface changes are
challenging to keep up with.

innocentkithinji:
It is easy to fix just change the googles daily.py file to look llike [this one](https://github.com/pydata/pandas-datareader/blob/master/pandas_datareader/google/daily.py).
The reason it doesnt fetch the data is that google changed its finance url from http://www.google.com/finance/historical to http://finance.google.com/finance/historical during the redirect the dates are not passed in the HTTP request.

EngJay:
I just ran into this issue with version 0.5.0 and confirmed that the fix suggested above by @innocentkithinji does restore expected functionality.

davidastephens:
Fixed in #404

Issue #394, opened 2017-09-15
Unable to fetch Google finance data


nasirali1:
Google has changed their URL from "http://www.google.com/finance/info" to "https://finance.google.com/finance"

rsvp:
OK, please tell us if and where this breaks the API.

What are the error messages with respect to pandas-datareader?

- [ ] Follow-up in near-duplicate issue #391

nasirali1:
Hi rsvp, 
Thanks for a prompt reply. Please find below the python code and error trace:

CODE:
def finance_data():
    """
    Stock indicies for
    .INX = S&P 500
    .DJI = Dow Jones
    .IXIC = NASDAQ
    """
    nds = web.get_quote_google(['.IXIC', '.DJI', '.INX'])

ERROR TRACE:
Traceback (most recent call last):
  File "/home/ali/IdeaProjects/Environmental-Data-Collection/core/finance/nds_stock.py", line 55, in <module>
    finance_date()
  File "/home/ali/IdeaProjects/Environmental-Data-Collection/core/finance/nds_stock.py", line 40, in finance_date
    nds = web.get_quote_google(['.IXIC', '.DJI', '.INX'])
  File "/home/ali/IdeaProjects/Environmental-Data-Collection/venv/lib/python3.5/site-packages/pandas_datareader-0.5.0-py3.5.egg/pandas_datareader/data.py", line 56, in get_quote_google
    return GoogleQuotesReader(*args, **kwargs).read()
  File "/home/ali/IdeaProjects/Environmental-Data-Collection/venv/lib/python3.5/site-packages/pandas_datareader-0.5.0-py3.5.egg/pandas_datareader/base.py", line 72, in read
    return self._read_one_data(self.url, self.params)
  File "/home/ali/IdeaProjects/Environmental-Data-Collection/venv/lib/python3.5/site-packages/pandas_datareader-0.5.0-py3.5.egg/pandas_datareader/base.py", line 79, in _read_one_data
    out = self._read_url_as_StringIO(url, params=params)
  File "/home/ali/IdeaProjects/Environmental-Data-Collection/venv/lib/python3.5/site-packages/pandas_datareader-0.5.0-py3.5.egg/pandas_datareader/base.py", line 90, in _read_url_as_StringIO
    response = self._get_response(url, params=params)
[...]

PR #404, opened 2017-10-01
Fix various failures in tests

davidastephens:


davidastephens:
resolves: #403 
resolves: #402 
fixes: #394 
fixes: #395

rsvp:
@davidastephens For the Google Finance test fix, the use of *pytest.skip* 
seems too definitive in the sense that Google may reconsider its API changes 
during the reconstruction period. So how about using *pytest.xfail* or even more 
explicitly `@pytest.mark.xfail()` with condition and reason attached?
See https://docs.pytest.org/en/latest/skipping.html

That way we can later run: `pytest --runxfail`
to see if Google has reverted without much hassle, 
and just remove the decorators.
(Historically, Google has reversed its pronouncements on quote 
retrieval via Sheets, so there's hope if there's enough public outcry :-).

rsvp:
hi @davidastephens -- could you kindly summarize what was fixed?

Many disparate issues were closed, so it might be useful to
***review the root problems which were solved***.

In what pandas_datareader version will the fixes be distributed? Thanks!

davidastephens:
Hi @rsvp: Each commit has detail as to what was fixed.  I aggregated into one pull request so I could get the travis tests to pass.

These fixes will be distributed in 0.6.0, but unfortunately, a significant portion of this package no longer works due to API removals.  I've noted which apis no longer work in the docs and skipped the tests, but unsure if we should do anything in the code.  Would adding DeprecationWarnings be helpful?

rsvp:
@davidastephens: hi, I think DeprecationWarnings would be most helpful 
since users would understand their own working code is not at fault, 
but rather API changes from the data vendors. The warnings should 
point to a summary page where the problems and limitations have 
been detailed, and possible remedies discussed. 

My best summary is here:  https://github.com/rsvp/fecon235/issues/7 
but when pandas_datareader 0.6.0 ships :shipit: I can write up a 
clean wiki page. I suggest  **pddr404api** as the hashtag keyword 
since that currently has no document matches via a Google search.

PR #413, opened 2017-10-22
TST: Changed skip to mark.xfail in test_google

davidastephens:
Changing skip to mark.xfail per recommendation of @rsvp in #404


rsvp:
@davidastephens hi, and ***thanks very much*** for clarifying the code.

PR #415, opened 2017-10-29
add __init__ otherwise it's not workiing[EOM]

treeguard:
- [x] tests passed

### DIFF
--- a/pandas_datareader/google/daily.py
+++ b/pandas_datareader/google/daily.py
@@ -29,6 +29,13 @@ class GoogleDailyReader(_DailyBaseReader):
         requests.sessions.Session instance to be used
     """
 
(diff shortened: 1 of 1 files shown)


davidastephens:
Thanks.  What fails without this?  The issue might be the URL needs updating (done in #404 )

pandas_datareader/google/daily.py
```diff
@@ -29,6 +29,13 @@ class GoogleDailyReader(_DailyBaseReader):
         requests.sessions.Session instance to be used
     """
 
+    def __init__(self, symbols=None, start=None, end=None, retry_count=3,
```
bashtage: This isn't necessary.  But if it were, the arguments should have the same order.

bashtage:
Closing since unlcear why this would be needed.

Issue #435, opened 2018-01-07
FRED no longer working


jhoodsmith:
FRED seems to have changed the URL structure for downloading CSV.

I'm using latest development version of pandas-datareader.

In using the example from the docs I get:

pandas_datareader._utils.RemoteDataError: Unable to read URL: http://research.stlouisfed.org/fred2/series/GDP/downloaddata/GDP.csv

oOo0oOo:
I can also confirm this bug. Getting the same Error.

Found another possible URL after a quick look but not sure if it works in all cases:
https://fred.stlouisfed.org/graph/fredgraph.csv?id=GDP

rsvp:
Cross-ref #436 
https://github.com/pydata/pandas-datareader/pull/436/files 

Thanks to @jhoodsmith

NumberChiffre:
Getting the same error and updated the url using fredgraph, thank you!

oOo0oOo:
Just a quick note: It seems the fred has reverted the URL; the original URL works again...

jhoodsmith:
Good news that the original is working again. 

It might be worth following point by @rsvp above, and make better use of qualifiers. I used 'id' only in #436, as this followed the original code. However, better use of 'freq' and/or begin/end dates will reduce impact on FRED servers.

jhoodsmith:
On a separate issue, I'm getting the same errors as reported by the #436 Travis build when I run tests locally.  There seems to general problems working with OECD and Google Options.

rsvp:
### Working status of old URL

So for the stlouisfed.org domain, how is the ***redirection***
between the *research* and *fred* subdomains taking place?

Any official announcements from FRED as to their intentions?
Perhaps it was an internal experiment accidentally brought 
into production.

bashtage:
Fixed

PR #436, opened 2018-01-07
New FRED URL

jhoodsmith:
Updated url to reflect the changes by fred.stlouisfed.org for downloading CSV.

### DIFF
--- a/pandas_datareader/fred.py
+++ b/pandas_datareader/fred.py
@@ -17,7 +17,7 @@ class FredReader(_BaseReader):
 
     @property
     def url(self):
-        return "http://research.stlouisfed.org/fred2/series/"
+        return "https://fred.stlouisfed.org/graph/fredgraph.csv"
 
     def read(self):
         try:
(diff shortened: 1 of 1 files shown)


bashtage:
This looks right and the FRED tests pass

bashtage:
Can you rebase on master?  It should pass and then this can be merged.

rsvp:
It is not clear whether the new URL schema at FRED is stable. 
For example, the default values on their qualifiers (see #435) 
may change without notice to pandas_datareader users, 
which will unexpectedly transform datasets.

The old URL is still operational, so it may be a good idea 
to wait to pull until revisions to the FRED API are made "official."

jhoodsmith:
For what it's worth, I agree with @rsvp. The new code is there if needed, but for the time being I suggest we follow the 'if it ain't broke, don't fix' strategy. Thank you to @bashtage for fixing the testing suit though.

bashtage:
The right way is https://research.stlouisfed.org/docs/api/fred/ but this requires an API key.  The graph endpoint has been around for years and is used extensively through FRED to power the graphs.

bashtage:
Accidental close.

bashtage:
Also worth noting that this new endpoint is where data comes from when one interactively downloads data.

bashtage:
I'm going to merge this.  I think it is the right endpoint since it is where data comes from when it is manually downloaded and there is no officially supported one.

PR #465, opened 2018-01-22
Added IEX Daily Reader (Historical)

addisonlynch:
Added IEX Daily Reader for historical data from IEX [chart](https://iextrading.com/developer/docs/#chart).

- Added tests in tests/test_iex_daily.py
- Added entry to docs/source/readers/remote_data.rst

Have an actions (Dividends+Splits) as well as quotes (from [quote](https://iextrading.com/developer/docs/#quote)) reader as well, but waiting for some clarification from the provider regarding bug fixes with the API calls - the quote may not be necessary anyway with the recent inclusion of "iex-last".

### DIFF
--- a/docs/source/remote_data.rst
+++ b/docs/source/remote_data.rst
@@ -29,6 +29,7 @@ extract data from various Internet sources into a pandas DataFrame.
 Currently the following sources are supported:
 
     - :ref:`Google Finance<remote_data.google>`
+    - :ref:`IEX<remote_data.iex>`
     - :ref:`Enigma<remote_data.enigma>`
     - :ref:`Quandl<remote_data.quandl>`
     - :ref:`St.Louis FED (FRED)<remote_data.fred>`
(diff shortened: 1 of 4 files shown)


rsvp:
@addisonlynch Thanks!

Docs, snippet: https://github.com/pydata/pandas-datareader/blob/master/docs/source/remote_data.rst#iex 

What are some limitations which should be noted? 
(Esp. from viewpoint of ex-users of yahoo-finance and google-finance.) 
Are ETFs and mutual funds supported? Queries/day? etc.

addisonlynch:
@rsvp 

Very similar to Yahoo/Google.

### Features
- List of available tickers is available from the [/ref-data/symbols](https://api.iextrading.com/1.0/ref-data/symbols)
- From what I tested a couple weeks back, the tickers are a superset of Nasdaq/NYSE/AMEX
- ETFs and most mutual funds supported. 


### Requirements
- No API key required
- No retrieval limits / throttling, but IEX does request that other readers (TOPS/DEEP) limit large request to 1/second

### Nuances
- Output is slightly different than Yahoo (does not include 'Adj Close')
- Data must be within a 5-year historical period (this is a limitation if migrating from Yahoo or Google)

rsvp:
@addisonlynch hi, and thanks for kind reply.
Updated https://github.com/rsvp/fecon235/issues/7#issuecomment-345898460
Many will be looking forward to your contributions in forthcoming 0.6.0 !

PR #488, opened 2018-02-15
Repaired Morningstar volume for indicies

### LINKED ISSUE #486: Morningstar queries to some tickers fail because the API returns no volumes
(I am a newbie at this, please guide me on how to contribute effectively)

Running
pandas_datareader.data.DataReader('HSI','morningstar',start=datetime.date(2012,12,22),end=datetime.date(2012,12,22))
fails because the Morningstar query for HSI does not return 'volumes' information.

This is the response directly from the API:
{"DividendData":[],"FairValueList":null,"Max":49000,"Min":9800,"PriceDataList":[{"DailyMaxNumPoint":-1,"Datapoints":[[29578.01,NaN,NaN,NaN]],"DateIndexs":[43089],"Description":"","Max":29578.01,"Min":29578.01,"Slot":3,"StartPos":29578.01,"Symbol":"XHKG:HSI","chartType":0}],"PriceType":"",**"VolumeList":null**}

I do not know how many tickers behave like this.

This specific case can be fixed by modifying pandas_datareader/mstar/daily.py as follows:
```
-        volumes = jsondata["VolumeList"]["Datapoints"]
+        try:
+            volumes = jsondata["VolumeList"]["Datapoints"]
+        except:
+            volumes = len(dateidx) * [ float('nan') ]

[...]

             if self.incl_vol is True:
-                bardict.update({"Volume": int(volumes[p] * 1000000)})
+                try:
+                    bardict.update({"Volume": int(volumes[p] * 1000000)})
+                except:
+                    bardict.update({"Volume": volumes[p]})
             else:
                 pass
```
There are other approaches - I do not know which one would be more aligned with the spirit of the module. Also, I have not investigated whether this change is safe in 100% of the use cases - it seems to be working fine for my project though.

Let me know if there is anything I can do to help!

addisonlynch:
Fixes #486. Certain Morningstar (index) tickers now use ``np.nan`` to indicate empty volume data. Thanks @ggomarr.

- [x] closes #xxxx
- [x] tests added / passed

### DIFF
--- a/pandas_datareader/mstar/daily.py
+++ b/pandas_datareader/mstar/daily.py
@@ -165,7 +165,7 @@ class MorningstarDailyReader(_BaseReader):
         if jsondata["VolumeList"]:
             volumes = jsondata["VolumeList"]["Datapoints"]
         else:
-            volumes = [np.nan] * 1000000
+            volumes = None
 
         dates = self._convert_index2date(indexvals=dateidx)
         barss = []
@@ -196,10 +196,10 @@ class MorningstarDailyReader(_BaseReader):
                     else:
                         pass
             if self.incl_vol is True:
-                try:
+                if volumes is None:
+                    bardict.update({"Volume": np.nan})
+                else:
                     bardict.update({"Volume": int(volumes[p] * 1000000)})
-                except ValueError:
-                    bardict.update({"Volume": volumes[p]})
             else:
                 pass
 
--- a/pandas_datareader/mstar/daily.py
+++ b/pandas_datareader/mstar/daily.py
@@ -4,6 +4,7 @@ from warnings import warn
 
 import requests
 from pandas import DataFrame
+import numpy as np
 
 from pandas_datareader._utils import SymbolWarning
 from pandas_datareader.base import _BaseReader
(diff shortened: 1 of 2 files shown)


addisonlynch:
Travis fails due to unrelated errors...since yesterday IEX Stock endpoints queries fail at random (Stats and Market data are OK). May want to add a warning for this-waiting for the provider's input first.

addisonlynch:
@bashtage can you try restarting the Travis build on this? The conda install errored in 0.19.2 for some reason and the unrelated provider issues should be resolved.

bashtage:
I restarted but it looked like there were failures against latest Pandas.

addisonlynch:
@bashtage thanks. Looks like the second round passed.

pandas_datareader/mstar/daily.py
```diff
@@ -161,7 +162,10 @@ def _restruct_json(self, symbol, jsondata):
 
         pricedata = jsondata["PriceDataList"][0]["Datapoints"]
         dateidx = jsondata["PriceDataList"][0]["DateIndexs"]
-        volumes = jsondata["VolumeList"]["Datapoints"]
+        if jsondata["VolumeList"]:
+            volumes = jsondata["VolumeList"]["Datapoints"]
+        else:
+            volumes = [np.nan] * 1000000
```
bashtage: Would be cleaner to set volumes to `None`

pandas_datareader/mstar/daily.py
```diff
@@ -192,7 +196,10 @@ def _restruct_json(self, symbol, jsondata):
                     else:
                         pass
             if self.incl_vol is True:
-                bardict.update({"Volume": int(volumes[p] * 1000000)})
+                try:
+                    bardict.update({"Volume": int(volumes[p] * 1000000)})
```
bashtage: Then chck if volume is not None...

bashtage:
Thanks

PR #490, opened 2018-02-19
ENH: Add support for AlphaVantage API

### LINKED ISSUE #389: Add Alpha Vantage as data source?
https://www.alphavantage.co/documentation/

Requires an API key

addisonlynch:
# AlphaVantage Readers

Adds support for nearly all of the AlphaVantage API endpoints.

## Additions

- **``AVTimeSeriesReader``** - class to handle requests for data from the Stock Time Series endpoints
  - top-level ``get_data_av``, for TIME_SERIES_DAILY
  - ``av-daily`` for TIME_SERIES_DAILY
  - ``av-daily-adjusted`` for TIME_SERIES_DAILY_ADJUSTED
  - ``av-weekly`` for TIME_SERIES_WEEKLY
  - ``av-weekly-adjusted`` for TIME_SERIES_WEEKLY_ADJUSTED
  - ``av-monthly`` for TIME_SERIES_MONTHLY
  - ``av-monthly-adjusted`` TIME_SERIES_MONTHLY_ADJUSTED
- **``AVQuotesReader``** - class to handle requests for Batch Stock Quotes using top-level ``get_quote_av``
- **``AVForexReader``** - class to handle FX Exchange Rates endpoint (``av-forex`` or top-level ``get_exchange_rate_av``)
- **``AVSectorPerformanceReader``** - class to handle Sector Performances SECTOR endpoint ``av-sector`` or top-level ``get_sector_performance_av``
- ``AlphaVantage``- base class for handling AlphaVantage requests

## Questions

- What format should the sector performances return? As of now, they're strings with a percent sign at the end. Using ``int64`` and ``float64`` for the other new readers.
- Is it okay for ``TimeSeriesReader`` to take the argument ``function`` which allows the selection of the daily/monthly/weekly endpoint? I have an implementation (see [this gist](https://gist.github.com/addisonlynch/91a921fa70c3a8fa647aeffcd562eb04)) which uses separate readers for each (``AVDailyReader``, ``AVDailyAdjustedReader``, etc.). Can switch over to that if it is more in-line with the current style.
- Used AV instead of AlphaVantage in names...readers such as ``AlphaVantageTimeSeriesReader`` seemed a bit long. 

## Needs Help

- **Will need to obtain an API key from AlphaVantage and contact them (support@alphavantage.co) to request an exemption from throttling to allow the tests to run.** Seems they don't like repeated calls to the same endpoints without additional permission. As such, the tests now will pass with an API key that does not have these restrictions, but may fail otherwise.
  - This key will be stored as the environment variable ``ALPHAVANTAGE_API_KEY``

## Near Future

The above takes care of most of the AlphaVantage endpoints. There are two groups remaining:

1. **Digital/Crypto Currencies** - is this something we would want to implement? I don't know if this is a direction you guys would want to head in for PDR. I don't have a lot of knowledge in this sphere, but it seems as if these endpoints could be easily implemented if desired.
2.  **Technical Indicators** - what is the preferred way to implement this?
  AlphaVantage provides 52 technical indicators, each of which seems to take the same default (required) parameters: ``function``, ``symbol``, ``interval``, ``apikey``. However, certain indicators (such as EMA) require additional parameters ``time_period``, ``series_type``, etc. 

In addition, will need to add a reader to handle TIME_SERIES_INTRADAY

- [X] closes #389 
- [X] tests added / passed
- [X] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [X] added entry to docs/source/whatsnew/vLATEST.txt

### DIFF
--- a/docs/source/remote_data.rst
+++ b/docs/source/remote_data.rst
@@ -181,6 +181,9 @@ symbols. The following endpoints are available:
                        access_key=os.getenv('ALPHAVANTAGE_API_KEY'))
     f.loc["2017-02-09"]
 
+The top-level function ``get_data_av`` is also provided. This function will
+return the ``TIME_SERIES_DAILY`` endpoint for the symbol and date range
+provided.
 
 Forex
 ^^^^^
--- a/docs/source/remote_data.rst
+++ b/docs/source/remote_data.rst
@@ -185,6 +185,25 @@ The top-level function ``get_data_av`` is also provided. This function will
 return the ``TIME_SERIES_DAILY`` endpoint for the symbol and date range
 provided.
 
+Quotes
+^^^^^^
+
+`AlphaVantage <https://www.alphavantage.co/documentation>`__ Batch Stock Quotes
+endpoint allows the retrieval of realtime stock quotes for up to 100 symbols at
+once. These quotes are accessible through the top-level function
+``get_quote_av``.
+
+.. ipython:: python
+
+    import os
+    from datetime import datetime
+    import pandas_datareader.data as web
+
+    web.get_quote_av(["AAPL", "TSLA"])
+
+
+.. note:: Most quotes are only available during market hours.
+
 Forex
 ^^^^^
 
--- a/docs/source/remote_data.rst
+++ b/docs/source/remote_data.rst
@@ -181,7 +181,8 @@ symbols. The following endpoints are available:
                        access_key=os.getenv('ALPHAVANTAGE_API_KEY'))
     f.loc["2017-02-09"]
 
-The top-level function ``get_data_av`` is also provided. This function will
+The top-level function ``get_data_alphavantage`` is also provided. This
+function will
 return the ``TIME_SERIES_DAILY`` endpoint for the symbol and date range
 provided.
 
--- a/docs/source/remote_data.rst
+++ b/docs/source/remote_data.rst
@@ -32,6 +32,7 @@ Currently the following sources are supported:
     - :ref:`Morningstar<remote_data.morningstar>`
     - :ref:`IEX<remote_data.iex>`
     - :ref:`Robinhood<remote_data.robinhood>`
+    - :ref:`AlphaVantage<remote_data.alphavantage>`
     - :ref:`Enigma<remote_data.enigma>`
     - :ref:`Quandl<remote_data.quandl>`
     - :ref:`St.Louis FED (FRED)<remote_data.fred>`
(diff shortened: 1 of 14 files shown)


gliptak:
@addisonlynch Please review the Travis failures

https://travis-ci.org/pydata/pandas-datareader/builds/343250680

While most of them API key related, this ones fails on ```import: pandas.testing```

https://travis-ci.org/pydata/pandas-datareader/jobs/343250681

@davidastephens @jreback Who can add the ```AlphaVantage``` key to ```Travis``` and ```ReadTheDocs```?

Thanks

bashtage:
Need to use same test pattern to skip module if API key is not available.

bashtage:
@gliptak I can add a key to Travis.  It won't actually be used for testing until the PR is merged since PRs can't access encryption variables to prevent leaking them by someone who could submit a useless PR that had `print(os.environ.get['SEKRIT'])`

addisonlynch:
I added the pytest.skipifs. Do you want me to add ``@skip_on_exception(RemoteDataError)`` to the tests as well? Certain tests will still fail with: 

>{'Information': 'Thank you for using the CURRENCY_EXCHANGE_RATE_API! Please kindly contact support@alphavantage.co if you would like to have a higher API call volume'}

unless PDR does obtain a higher call volume key. Let me know what works best.

docs/source/remote_data.rst
```diff
@@ -145,6 +146,106 @@ year relative to today.
...
+    import os
+    from datetime import datetime
+    import pandas_datareader.data as web
+
+    f = web.DataReader("AAPL", "av-daily", start=datetime(2017, 2, 9),
+                       end=datetime(2017, 5, 24),
+                       access_key=os.getenv('ALPHAVANTAGE_API_KEY'))
+    f.loc["2017-02-09"]
+
+The top-level function ``get_data_av`` is also provided. This function will
```
bashtage: I don't really like get_data_av -- how could a new user know what av is?  Maybe just the long get_data_alphavantage? Autocomplete is pretty good.

pandas_datareader/av/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+from pandas_datareader._utils import RemoteDataError
+
+import pandas as pd
+
+AV_BASE_URL = 'https://www.alphavantage.co/query'
+
+
+class AlphaVantage(_BaseReader):
+    """
+    Base class for all AlphaVantage queries
```
bashtage: I think there should be docstring here.  Maybe we should use a doc string inheriter to simplify this?

pandas_datareader/av/forex.py
```diff
@@ -0,0 +1,93 @@
...
+class AVForexReader(AlphaVantage):
+    """
+    Returns DataFrame of the AlphaVantage Foreign Exchange (FX) Exchange Rates
+    data.
+
+    .. versionadded:: 0.7.0
+
+    Parameters
+    ----------
+    pairs : string, array-like object (list, tuple, Series)
```
bashtage: Why not just keep the standard symbols?

pandas_datareader/av/__init__.py
bashtage: Nevermind - this is a base class.

pandas_datareader/av/sector.py
```diff
@@ -0,0 +1,39 @@
...
+
+class AVSectorPerformanceReader(AlphaVantage):
+    """
+    Returns DataFrame of the AlphaVantage Sector Performances SECTOR data.
+
+    .. versionadded:: 0.7.0
+
+    Parameters
+    ----------
+    retry_count : int, default 3
```
bashtage: I don't think this docstring has all the arguments from the base class.

pandas_datareader/data.py
```diff
@@ -47,6 +51,10 @@
            'DataReader']
 
 
+def get_data_av(*args, **kwargs):
```
bashtage: get_data_alphavantage?

addisonlynch:
Done.

addisonlynch:
Done

addisonlynch:
@bashtage was going to add the remaining endpoints today or tomorrow. Do you want me to push them to this pull or a separate one? Also, can you provide any direction on the preferred way to implement the technical indicators (see above). Thanks

gliptak:
@bashtage Were your review items addressed?

@addisonlynch Could you resolve the conflict?

Thanks

addisonlynch:
@gliptak conflicts resolved. Did indeed repair the requested changes

addisonlynch:
Travis fails on unrelated errors (FRED and tiingo?). Are we ready to merge this otherwise?

gliptak:
@bashtage Did you complete your review? Thanks

Issue #376, opened 2017-08-03
Add Enigma Public Data Connector


hydrosquall:
The Enigma reader currently points to `app.enigma.io`, which will soon be shut down due to the recent release of `public.enigma.com` on June 20, 2017.  Writeup is [here](https://www.enigma.com/blog/the-new-enigma-public).

In this new API, the concept of "datapaths" has been removed, and replaced with unique `uuid4` identifiers for every dataset.

I will base a new connector on the previous one, and recommend that the old connector be removed when `app.enigma.io` is officially shut down.

gliptak:
https://github.com/pydata/pandas-datareader/issues/370

gliptak:
@hydrosquall Consider closing this issue.

hydrosquall:
Thanks @gliptak, you're right!

PR #499, opened 2018-03-07
Allows user to input a custom base url to EnigmaReader

EmilRex:
Adds support for optionally passing a custom `base_url` to the `EnigmaReader`. The default behavior will not change, but this will allow the reader to be used with dev/staging instances as well as custom deployments. The feature will also guard against any further domain changes such as in #376.

### DIFF
--- a/pandas_datareader/enigma.py
+++ b/pandas_datareader/enigma.py
@@ -13,6 +13,22 @@ class EnigmaReader(_BaseReader):
     Collects current snapshot of Enigma data located at the specified
     data set ID and returns a pandas DataFrame.
 
+    Parameters
+    ----------
+    dataset_id : str
+        Enigma dataset UUID.
+    api_key : str, optional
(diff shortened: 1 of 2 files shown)


EmilRex:
@bashtage This is ready for review. Happy to fix anything I might have left out and really appreciate the time. :bowtie:

pandas_datareader/enigma.py
```diff
@@ -32,7 +32,8 @@ def __init__(self,
                  api_key=None,
                  retry_count=5,
                  pause=.75,
-                 session=None):
+                 session=None,
+                 base_url="https://public.enigma.com/api"):
```
bashtage: Needs something added to the docstring.  Ideally, the docstring would be rewritten to include all parameters.

EmilRex: @bashtage Added a `Parameters` section to the `EnigmaReader` docstring. Let me know what you think.

PR #514, opened 2018-04-08
FIX: Morningstar Retry Issue

### LINKED ISSUE #513: morningstar reader does not increment retry count
I have a program that is failing on two symbols. AAXN and AABA. It creates a stack overflow error eventually.

As far as I can tell it's because retry count never gets incremented. And then it recursively calls itself to try again. 

https://github.com/pydata/pandas-datareader/blob/master/pandas_datareader/mstar/daily.py

https://stackoverflow.com/questions/49703418/stack-overflow-with-pandas-datareader/49705013#49705013

Also, can you explain to me why it appears that the if else is entirely contained within the exception.

dtemkin:
symbol_data now carries over in successive recursions of _dl_mult_symbols(). Also eliminated some unused objects and redundancies.

- [x] closes #513 
- [ ] tests added / passed

### DIFF
--- a/pandas_datareader/mstar/daily.py
+++ b/pandas_datareader/mstar/daily.py
@@ -62,8 +62,6 @@ class MorningstarDailyReader(_BaseReader):
         self.currency = currency
         self.interval = interval
 
-        self._symbol_data_cache = []
-
     def _url_params(self):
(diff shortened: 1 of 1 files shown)


dtemkin:
Build is failing due to Google test configuration.

PR #515, opened 2018-04-08
Morningstar Retry Fix and Google API Patch

### LINKED ISSUE #502: New problems with Google Finance?
I got the following traceback/error:

> raise RemoteDataError('Unable to read URL: {0}'.format(url))
> pandas_datareader._utils.RemoteDataError: Unable to read URL: 
> http://finance.google.com/finance/historical?q=ETR%3AADS&startdate=Jan+01%2C+2000&enddate=Mar+16%2C+2018&output=csv

Does others have the same problem? (Thanks.)

Are they joking?

> We're sorry...
> ... but your computer or network may be sending automated queries. To protect our users, we can't process your request right now.
>
> See Google Help for more information.

Have you the same problem? How do you solve this problem?

### LINKED ISSUE #513: morningstar reader does not increment retry count

dtemkin:
- [x] closes #513 
- [x] closes #502 
- [x] tests added / passed

### DIFF
--- a/docs/source/remote_data.rst
+++ b/docs/source/remote_data.rst
@@ -29,6 +29,7 @@ extract data from various Internet sources into a pandas DataFrame.
 Currently the following sources are supported:
 
     - :ref:`Google Finance<remote_data.google>`
+    - :ref:`Tiingo<remote_data.tiingo>`
     - :ref:`Morningstar<remote_data.morningstar>`
     - :ref:`IEX<remote_data.iex>`
     - :ref:`Robinhood<remote_data.robinhood>`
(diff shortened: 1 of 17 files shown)


dtemkin:
I also just implemented a 'stop-gap' fix for the failing google finance api. The biggest thing was changing the GoogleDailyReader url to 'https://finance.google.co.uk/bctzjpnsun/historical'

PR #524, opened 2018-04-30
Fix error generated by missing keys

stoffprof:
This fixes a problem where errors are generated if requesting data for a ticker that has no corporate actions data.


bashtage:
I renamed FX.py to fx.py to reflect common practice of always using lower in file names.  Could you change this so that it will merge?

If you are using Windows, this can be a pain since it isn't case sensitive.  You need to do something like

```
ren FX.py temp.py
ren temp.py fx.py
```

bashtage:
It looks like you need to rebase on the master copy here, since many of these patches were previously merged.

bashtage:
Something like

```
git remote add upstream https://github.com/pydata/pandas-datareader
git fetch --all
git rebase upstream/master
```
[prepare for pain]

The pain happens if you have files that are in conflict with the master.  In which case you will get blocks that looks like

<<<<<<
CODE FROM MASTER
================
YOUR CODE
>>>>

You have to find all of these blocks and decide what you think should be here.  Once you have "resolved" these, you need to do

```
git add /path/to/file/fixed.py
```

for any file that was in conflict, and then

``` 
git rebase --continue
```

bashtage:
See 

https://wincent.com/wiki/Understanding_Git_conflict_markers

bashtage:
I got it backward, top if your code, bottom is from upstream.

stoffprof:
You weren't kidding about the pain. I'm not sure exactly what I just did, but I *think* it's ready to go now. Let me know if I need to do something else.

davidastephens:
I rebased this and also reinstated the yahoo-actions API in #525.  I also returned the daily output to a dataframe rather than a dict so its no longer a breaking change.

Please let me know if you have any comments before I merge.

FYI - there is a new api for quotes on yahoo as well, so yahoo quotes can be reimplemented.  I'll try to create a pull request for that in the next week or so if no one does it first.

https://query1.finance.yahoo.com/v7/finance/quote?symbols=AAPL

PR #525, opened 2018-05-12
Reinstate Yahoo API

davidastephens:
- [X] closes #524 
- [X] tests added / passed
- [X] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [X ] added entry to docs/source/whatsnew/vLATEST.txt

PR #529, opened 2018-05-25
test on 0.23.0

jreback:


addisonlynch:
The test cases which fail here are repaired in @davidastephens #525.

PR #560, opened 2018-08-03
Update fred.py

Willseed:
Because the is_list_like is moved to pandas.api.types, I change the fred.py file which is highlighted in the picture. I replace from pandas.core.common import is_list_like with from pandas.api.types import is_list_like, and it works.

reference：https://stackoverflow.com/questions/50394873/import-pandas-datareader-gives-importerror-cannot-import-name-is-list-like

- [ ] closes #xxxx
- [ ] tests added / passed

### DIFF
--- a/pandas_datareader/fred.py
+++ b/pandas_datareader/fred.py
@@ -1,4 +1,4 @@
-from pandas_datareader.compat import is_list_like
+from pandas.api.types import is_list_like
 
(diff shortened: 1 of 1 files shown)


bashtage:
The current import is correct, `from pandas_datareader.compat import is_list_like`.  This is a compatibility issue.

PR #564, opened 2018-08-13
Fixed a typo in an exception message

Mottl:


Mottl:
@bashtage, if this PR is too unclear, please close — I will create another one with a single commit.

PR #565, opened 2018-08-13
Fixed an issue with passing pandas DataFrame as a parameter

Mottl:
Fixed an issue with passing pandas DataFrame object as a `symbols` parameter for Moex data connector:

```python
import pandas as pd

symbols = pd.DataFrame(index=['SBER', 'USD000UTSTOM'])
df = web.DataReader(symbols, 'moex',
                    start='2017-07-01', end='2017-07-31')
```

### DIFF
--- a/pandas_datareader/moex.py
+++ b/pandas_datareader/moex.py
@@ -11,32 +11,32 @@ from pandas_datareader.compat import is_list_like
 
 class MoexReader(_DailyBaseReader):
     """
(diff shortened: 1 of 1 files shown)


bashtage:
Could you add a whats new for this bug?

Mottl:
There is a statement in _whatsnew_ already about multiple symbols in a constructor:

    - MOEX data connector now supports multiple symbols in a constructor.

I doubt we should add something more, because this is just a patch for this topic.

PR #591, opened 2018-10-20
ENH: Bug fix for downloading index historicals.

corakwue:
`import pandas_datareader.data as web
f = web.DataReader('^DJI', 'stooq')`

Above code breaks because country code is pre-appended for any symbol. For index data (^dji, ^spx etc), there is no need to pre-append country code. This patch fixes that.

- [ ] closes #xxxx
- [ ] tests added / passed

### DIFF
--- a/pandas_datareader/stooq.py
+++ b/pandas_datareader/stooq.py
@@ -37,12 +37,13 @@ class StooqDailyReader(_DailyBaseReader):
 
     def _get_params(self, symbol, country='US'):
         symbol_parts = symbol.split(".")
-        if len(symbol_parts) == 1:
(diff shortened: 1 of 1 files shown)


corakwue:
I checked, the CI build failures appear to be unrelated to the commit.

bashtage:
Could you add a test that uses symbols with all patterns covered by the if then block?

corakwue:
> Could you add a test that uses symbols with all patterns covered by the if then block?

Done in #592

addisonlynch:
@bashtage could you merge these stooq PRs if your review is complete? Lots of issues coming up here and on stack in regards to this problem.

PR #601, opened 2018-12-21
Fix Yahoo Daily Reader

jmgate:
Some time ago (~2 years, maybe?), Yahoo decided to get rid of its historical price data API.  When that happened, `pandas-datareader`'s daily Yahoo reader ceased to work.  This PR gets it back up and running again.  Yahoo hasn't released a new API, and it seems they don't intend to, but you can still get at the data in CSV format through a "Download Data" link on a security's [Historical Data](https://finance.yahoo.com/quote/AMZN/history?p=AMZN) page.  These changes use that URL to provide the same historical data we all knew and loved.

Additionally, there are other hoops to jump through involving getting the Yahoo Finance cookie jar and associated cookie crumb to allow communication between you and the service.  This is automated behind the scenes so the user is pleasantly oblivious.

- [ ] closes #xxxx
- [ ] tests added / passed

### DIFF
--- a/pandas_datareader/base.py
+++ b/pandas_datareader/base.py
@@ -113,7 +113,7 @@ class _BaseReader(object):
         """
         return response.content
 
-    def _get_response(self, url, params=None, headers=None):
+    def _get_response(self, url, params=None, headers=None, cookies=None):
         """ send raw HTTP request to get requests.Response from the specified url
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


jmgate:
Running `pytest` now&hellip;

jmgate:
Looks like I've got two failing Yahoo tests:
*  `TestYahoo.test_get_data_yahoo_actions`
*  `TestYahoo.test_yahoo_DataReader`

Both of these are also failing on `master`.  Things are behaving differently on my branch, but that because I don't think this CSV method of querying the historical price data supports actions.  Perhaps it's possible to re-release the Yahoo daily reader with the caveat that no action information is available.

davidastephens:
Yahoo historical and actions are currently working in master.

PR #602, opened 2018-12-26
Fixed two dividend related issues

peijiche:
-Issue 1:  DataReader(symbol, 'yahoo-dividends', start, end) will have divided by zero error for some dividend stocks like "NTR". 

Root cause:  when the stock has event like "1/0 Stock Split", it will fail. 

The fix: do numerator check first before evaluating the split ratio. In the above case, it should be Splitratio=1

-Issue 2:  fixed a bug that failed dividend/split test case of "def test_yahoo_DataReader(self)" in tests/test_yahoo.py.

### DIFF
--- a/pandas_datareader/tests/yahoo/test_yahoo.py
+++ b/pandas_datareader/tests/yahoo/test_yahoo.py
@@ -278,6 +278,24 @@ class TestYahoo(object):
         exp.index.name = 'Date'
         tm.assert_frame_equal(result.reindex_like(exp).round(5), exp.round(5))
 
+
+        # test cases with "1/0" split ratio in actions- no split, just chnage symbol from POT to NTR
+        start = datetime(2017, 12, 30)
+        end = datetime(2018, 12, 30)
+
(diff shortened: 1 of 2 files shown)


addisonlynch:
Can you please add relevant tests for this issue? Thanks!

peijiche:
Thanks for the reply. Those two are bug fixes and are covered already by the existing test cases. I could think about adding one or more if needed.
Happy New Year!
-Peiji
    On Monday, December 31, 2018, 6:19:05 AM PST, Addison Lynch <notifications@github.com> wrote:  
 
 
Can you please add relevant tests for this issue? Thanks!

peijiche:
added one test case for symbol NTR. It was failed before with divided by zero error.

peijiche:
Happy New Year to you too!
Peiji
    On Monday, December 31, 2018, 11:23:24 AM PST, Addison Lynch <notifications@github.com> wrote:  
 
 
@peijiche perfect...I should have been more specific, but that's exactly what I was referring to. Thank you and happy new year!

cc @bashtage

Issue #623, opened 2019-03-02
IEX Cloud as new data source?


gliptak:
Next iteration of IEX data

https://iexcloud.io/

https://iextrading.com/developer/docs/

addisonlynch:
@gliptak thanks...I was planning to roll my implementation from [iexfinance](https://github.com/addisonlynch/iexfinance) into this (I wrote the original IEX daily reader for PDR). Most of the endpoints have remained the same, but there are a few new ones, including financial statements, which we may want to implement.

IEX is maintaining their old API until at least June, so I was planning on using that for PDR until then, as it does not require registration and is free.

gliptak:
yes, we are to obtain a token to be used for ```Travis```

addisonlynch:
See https://iexcloud.io/docs/api/#sandbox

kabonker:
So will we all have to get like API keys and accounts?   Will the current Pandas datareader for IEX be upgraded to a new one?   Or will it just end?  June is fast approaching, I think they give the data back as JSON but have not tried to interface yet.

addisonlynch:
@kabonker yes. We will port the existing readers to the IEX Cloud platform. Legacy V1 (no auth) API will be discontinued by the provider on 6/1/19, so it will be done by then. You will need to sign up for an API key, as IEX Cloud requires authentication.

gliptak:
Closing with https://github.com/pydata/pandas-datareader/pull/638 merged

PR #638, opened 2019-05-29
COMPAT: Migrate IEX Daily Reader to IEX Cloud

addisonlynch:
As discussed in #623, IEX is migrating to a new platform, [IEX Cloud](https://iexcloud.io) on June 1, 2019. At this time, support for the legacy API (existing source for PDR's IEX readers) will end.

This PR migrates ``IEXDailyReader`` to IEX Cloud. Additional work to follow will migrate the remainder of the readers.

IEX provides a sandbox environment, which must be enabled by setting ``IEX_SANDBOX`` to ``enable`` to run tests. A test API key must be used (from IEX Cloud console). 

- [ ] closes #xxxx
- [x] tests added / passed


addisonlynch:
@bashtage will probably want to cut a release for this. Seems to be one of the higher-usage readers and will break when the legacy platform is continued on 6/1. I'll see if I can migrate the rest of the readers before then.

bashtage:
Any idea what is actually broken?  And is it fixable or does it need immediate deprecation

addisonlynch:
IEX is migrating off of their legacy API (a free, no auth service) to IEX Cloud which requires auth. URL changes but functionality is the same. This reader will break on 6/1...don't see a need to deprecate and create a new one and add auth. Adding auth to this reader should be fine with a few other smaller modifications. I realized I need to clean up one small caveat with the test suite here then will be good to go.

addisonlynch:
@bashtage can you review & merge? IEX readers are now dead from provider as is

jzstoller:
Any idea when this might be fixed? thanks!

PR #648, opened 2019-06-26
Fixed environment variable name

oceanapplications:


addisonlynch:
Probably should have stuck with the previous naming convention using ``_API_KEY`` as the suffix. @bashtage will likely decide if this should change given that the current convention is now merged into dev but has not been released. Either ``IEX_API_TOKEN`` or ``IEX_API_KEY`` works regardless.

However, use of this package does *not* require the publishable token and the value does not need to be set for the documentation, as a key will be added to the travis environment, in which the documentation is built by doctr.

Thanks!

bashtage:
Could you fix it in what's new as well?

bashtage:
Thanks.

PR #649, opened 2019-07-02
Limit data use

oceanapplications:
Currently the minimum requested data is a full year.  This limits the request to more specific time frames.

The current functionality is undesirable because IEX is no longer unlimited.  After this change, for less than 6 days,  the cost is 50 messages.  Before it was over 3,000.

- [x] tests added / passed

### DIFF
--- a/pandas_datareader/iex/daily.py
+++ b/pandas_datareader/iex/daily.py
@@ -101,6 +101,16 @@ class IEXDailyReader(_DailyBaseReader):
         elif 1 <= years < 2:
             return "2y"
         elif 0 <= years < 1:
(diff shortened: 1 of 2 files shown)


addisonlynch:
LGTM..thanks!

bashtage:
Needs a test.

bashtage:
It is fine to directly test private functions internally.  Private doesn't mean private to pdr, only the everything else.

bashtage:
In particular all of the new paths need to be visited.

bashtage:
Need to get some coverage reporter attached

bashtage:
Thanks.

PR #664, opened 2019-08-27
COMPAT: Repair PY2 issues

### LINKED ISSUE #665: Not working with new pandas versions
When i'm trying install `pandas-datareader` both with `pandas >= 0.25.0`
i getting error like `AttributeError: module 'pandas.compat' has no attribute 'binary_type'` when trying loading data from MOEX for example.
I'm using `pandas==0.24.2` and all working fine. Please, fix this issue!

addisonlynch:
Merging 0.7.4 py2 compat changes into master in prep for 0.8.0

- [x] closes #665
- [ ] tests added / passed

### DIFF
--- a/docs/source/whatsnew/v0.7.0.txt
+++ b/docs/source/whatsnew/v0.7.0.txt
@@ -64,6 +64,7 @@ Backwards incompatible API changes
 Bug Fixes
 ~~~~~~~~~
 
+- Fixed import of pandas.compat (:issue:`657`)
 - Added support for passing the API KEY to QuandlReader either directly or by
   setting the environmental variable QUANDL_API_KEY (:issue:`485`).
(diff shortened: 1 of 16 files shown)


bashtage:
The important question is whether pdr should be Python 3 only for 0.8 on, in which case all of the PY2/3 compat code can be removed

bashtage:
Either 0.8 is the last PY2 compat release of 0.7 is the last PY2 release.  If it is the latter, then we can remove all PY2/3 code now.

bashtage:
Need to add 3.7 for testing and get tests/xfail before release.

gliptak:
With build green, could 0.8.0pre be tagged?

bashtage:
I think docs need to be updated before the tag, but close.

bashtage:
FWIW current master is now correctly tagged 0.8.0dev0

Issue #662, opened 2019-08-25
AlphaVantage raises "Please input a valid date range" for previously working code


schneiderfelipe:
Hi all,

The following snippet was working fine until I updated pandas-datareader:

```python3
start = datetime.datetime(1990, 5, 16)
end = datetime.datetime.now()
data = web.DataReader("ITUB3.SAO", "av-daily-adjusted", start=start, end=end)
```

Now it throws a `ValueError` ("Please input a valid date range"). Even leaving `start` and `end` as `None` doesn't work.

I believe it has to do with the indexing at line 81 of `pandas_datareader.av.time_series`:

https://github.com/pydata/pandas-datareader/blob/8eb5655ff90ed043f9297b9585c8ca2528f5c84c/pandas_datareader/av/time_series.py#L81

What do you think? Thanks in advance.

dahifi:
Looks like it's still an issue @addisonlynch. Using the following code from the documentation: 
``f = web.DataReader("AAPL", "av-daily", 
                   start=datetime(2017, 2, 9), 
                   end=datetime(2017, 5, 24), 
                   api_key=os.getenv('ALPHAVANTAGE_API_KEY')
                  )

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-14-fd942bc63d60> in <module>
      2                    start=datetime(2017, 2, 9),
      3                    end=datetime(2017, 5, 24),
----> 4                    api_key=os.getenv('ALPHAVANTAGE_API_KEY')
      5                   )

~/.local/share/virtualenvs/tradePlan-24nzYghD/lib/python3.5/site-packages/pandas/util/_decorators.py in wrapper(*args, **kwargs)
    186                 else:
    187                     kwargs[new_arg_name] = new_arg_value
--> 188             return func(*args, **kwargs)
    189         return wrapper
    190     return _deprecate_kwarg

~/.local/share/virtualenvs/tradePlan-24nzYghD/lib/python3.5/site-packages/pandas_datareader/data.py in DataReader(name, data_source, start, end, retry_count, pause, session, api_key)
    577             pause=pause,
    578             session=session,
--> 579             api_key=api_key,
    580         ).read()
    581 

~/.local/share/virtualenvs/tradePlan-24nzYghD/lib/python3.5/site-packages/pandas_datareader/base.py in read(self)
     98         """Read data from connector"""
     99         try:
--> 100             return self._read_one_data(self.url, self.params)
    101         finally:
[...]

Issue #582, opened 2018-09-20
AlphaVantage 5/minute ratelimit is making tests fail


bashtage:
We would need up to ~50 minute to pass tests.  Also affect doc building.

Should switch to allowed fail on RemoteData error with the right text.  Should also return this as part of the RemoteDataError

`Thank you for using Alpha Vantage! Please visit https://www.alphavantage.co/premium/ if you would like to have a higher API call volume`

bashtage:
I don't think we should stretch it out.  We could do something more clever like (a) only testing in 1 config (easy) and (b) choosing a random subset of 5 tests to run, so that on average we get complete coverage eventually. 

The other alternative would be to approach AV to ask for an improved key for testing only.

gliptak:
it doesn't stretch it out that long, sort of processes in bursts. I will submit a PR for this and https://github.com/pydata/pandas-datareader/issues/673 together.

bashtage:
How are you going to deal with the 5 concurrent jobs on Travis?

gliptak:
@bashtage ok, to that point the value might need to be higher. The good news is that this doesn't run PRs ...

bashtage:
Maybe you could enable it for one job.

bashtage:
Fixed as much as it can be.

PR #677, opened 2019-09-14
Correct AlphaVantage data source

### LINKED ISSUE #673: AlphaVantage inconsistent index handling for timeseries
API documentation describes index values as ```string```

https://github.com/pydata/pandas-datareader/blob/master/pandas_datareader/av/time_series.py#L15

while processes as ```datetime```

https://github.com/pydata/pandas-datareader/blob/master/pandas_datareader/av/time_series.py#L85

The unit test sends it in as ```datetime```

https://github.com/pydata/pandas-datareader/blob/master/pandas_datareader/tests/av/test_av_time_series.py#L41

but validates results as ```string```

https://github.com/pydata/pandas-datareader/blob/master/pandas_datareader/tests/av/test_av_time_series.py#L46

What would be the preferred approach? Updating either way might break existing code ...

https://github.com/pydata/pandas-datareader/issues/582
https://github.com/pydata/pandas-datareader/issues/662

gliptak:

https://travis-ci.com/gliptak/pandas-datareader/jobs/235398520

```
pandas_datareader/tests/av/test_av_forex.py list index out of range
.list index out of range
...
pandas_datareader/tests/av/test_av_quotes.py ....x
pandas_datareader/tests/av/test_av_sector.py .
pandas_datareader/tests/av/test_av_time_series.py ........
```

- [ ] closes #673 #582 #662 
- [ ] tests added / passed

### DIFF
--- a/pandas_datareader/av/__init__.py
+++ b/pandas_datareader/av/__init__.py
@@ -60,7 +60,7 @@ class AlphaVantage(_BaseReader):
         except KeyError:
             if "Error Message" in out:
                 raise ValueError("The requested symbol {} could not be "
-                                 "retrived. Check valid ticker"
+                                 "retrieved. Check valid ticker"
                                  ".".format(self.symbols))
             else:
                 raise RemoteDataError()
--- a/pandas_datareader/tests/av/test_av_forex.py
+++ b/pandas_datareader/tests/av/test_av_forex.py
@@ -32,7 +32,7 @@ class TestAlphaVantageForex(object):
     @pytest.mark.skipif(TEST_API_KEY is None,
                         reason="ALPHAVANTAGE_API_KEY not set")
     def test_one_pair(self):
-        df = web.DataReader("USD/EUR", "av-forex")
+        df = web.DataReader("USD/EUR", "av-forex", retry_count=6, pause=20.5)
         assert isinstance(df, pd.DataFrame)
         assert df.loc["To_Currency Name"][0] == "Euro"
         assert df.loc["Time Zone"][0] == 'UTC'
@@ -41,6 +41,6 @@ class TestAlphaVantageForex(object):
                         reason="ALPHAVANTAGE_API_KEY not set")
     def test_multiple_pairs(self):
         pairs = ["USD/JPY", "EUR/JPY"]
-        df = web.DataReader(pairs, "av-forex")
+        df = web.DataReader(pairs, "av-forex", retry_count=6, pause=20.5)
         assert isinstance(df, pd.DataFrame)
         assert df.columns.equals(pd.Index(pairs))
--- a/pandas_datareader/tests/av/test_av_quotes.py
+++ b/pandas_datareader/tests/av/test_av_quotes.py
@@ -33,7 +33,7 @@ class TestAVQuotes(object):
     @pytest.mark.skipif(TEST_API_KEY is None,
                         reason="ALPHAVANTAGE_API_KEY not set")
     def test_single_symbol(self):
-        df = web.get_quote_av("AAPL")
+        df = web.get_quote_av("AAPL", retry_count=6, pause=20.5)
         assert len(df) == 1
 
         expected = pd.Index(["price", "volume", "timestamp"])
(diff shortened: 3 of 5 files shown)


pandas_datareader/av/time_series.py
```diff
@@ -13,11 +14,10 @@ class AVTimeSeriesReader(AlphaVantage):
     ----------
     symbols : string
         Single stock symbol (ticker)
-    start : string, (defaults to '1/1/2010')
-        Starting date, timestamp. Parses many different kind of date
-        representations (e.g., 'JAN-01-2010', '1/1/10', 'Jan, 1, 1980')
-    end : string, (defaults to today)
-        Ending date, timestamp. Same format as starting date.
+    start : pd.Timestamp
```
bashtage: These need to match _BaseReader, which is ultimately at the end of a super chain. `str` should always be accepted even if `TimeStamp` is used internally rather than `datetime.datetime`.

gliptak: Many existing methods don't seem to follow:

https://github.com/pydata/pandas-datareader/blob/master/pandas_datareader/tsp.py#L54
https://github.com/pydata/pandas-datareader/blob/master/pandas_datareader/iex/daily.py#L95
https://github.com/pydata/pandas-datareader/blob/master/pandas_datareader/eurostat.py#L23

I'm not sure what would be a good update here ...

gliptak: @bashtage  @addisonlynch how would you like to see this moving forward?

bashtage:
The big question is what are the acceptable formats fot start and end?  Is it just TimeStamp? or is it TimeStamp, datetime and str. I think the latter makes more sense. Of course, ideally this would be universal.

gliptak:
OK, so If I go back to ```string``` for ```start/end```, do you see value in ```retry_count/pause``` updates?

PS Can I still use Travis to unstable tests after https://github.com/pydata/pandas-datareader/pull/682 ?

bashtage:
Yes, I can see the value of pause here, and retry.  

The plan for AV will be to split it into its own container and so it will run exactly once on travis and so it should be possible to not violate the rate limit.

I was thinking of splitting some others of the more problematic reader or into their own run so that is is easier to see when a single reader is breaking, and to move it to OK fail.  The current list is just a first guess.

addisonlynch:
I think we need to revisit ``_sanitize_dates`` and decide which formats that helper should allow. Run every reader through this helper, which should accept str, int, TimeStamp, DateTime, Date (?).

https://github.com/pydata/pandas-datareader/blob/4c86c94aca1e07a1d64d768da1fa37579ca6490a/pandas_datareader/_utils.py#L17-L38

If there are nuances for a certain reader (i.e. different default date ranges), then these should be checked by *that reader* before calling ``_sanitize_dates``. In other words, this should be a universal helper (also importantly checks start < end).

addisonlynch:
@bashtage for CI I think that is a good plan...if we can make it a build which can accept other readers that would be good too. IEX now rate limits as well and if we ever convert the other IEX readers here to IEX cloud, they will face the same problems as AV.

addisonlynch:
Only other thought I had for CI is how is all of this going to effect the docbuild? Are we going to need the pause values above for everything to finish?

gliptak:
@bashtage  I pushed ```retry_count/pause``` updates only

bashtage:
Thanks.

bashtage:
> Only other thought I had for CI is how is all of this going to effect the docbuild?

Docs should be modified to not run any code that depends on endpoints that require an API key so that they can be built by anyone.

bashtage:
> If there are nuances for a certain reader (i.e. different default date ranges), then these should be checked by _that reader_ before calling `_sanitize_dates`. In other words, this should be a universal helper (also importantly checks start < end).

Or `sanitize_dates` could be extended to allow optional inputs like `earliest` or `latest` so that start>=earliest and end<latest (pythonic equality structure, start:end, but latest could also be <=, not that I think anyone would use latest.), if these are not None.

PR #679, opened 2019-09-15
Reverse data after slice.

### LINKED ISSUE #662: AlphaVantage raises "Please input a valid date range" for previously working code

jhoodsmith:
Closes #662 by fixing #662.

The problem with #662 was that the reversing of the data happened before the `data.loc[start_str:end_str]`. Unless `start_str` = `end_str`, this would always return empty.

I could have equally made the same fix by changing `data.loc[start_str:end_str]` to `data.loc[end_str:start_str]`.

### DIFF
--- a/pandas_datareader/av/time_series.py
+++ b/pandas_datareader/av/time_series.py
@@ -80,11 +80,13 @@ class AVTimeSeriesReader(AlphaVantage):
 
     def _read_lines(self, out):
         data = super(AVTimeSeriesReader, self)._read_lines(out)
-        # reverse since alphavantage returns descending by date
-        data = data[::-1]
         start_str = self.start.strftime('%Y-%m-%d')
         end_str = self.end.strftime('%Y-%m-%d')
         data = data.loc[start_str:end_str]
+
+        # reverse since alphavantage returns descending by date
+        data = data[::-1]
+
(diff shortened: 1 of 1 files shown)


gliptak:
@jhoodsmith it seems that the Aphavantage time-series API was reset #677

jhoodsmith:
In that case we can solve by revoking #662 and ignoring this pull (#679) - i.e., by removing the line `data = data[::-1]`.

bashtage:
Going to close.  Can reopen if this is still an issue.  Thanks.

PR #683, opened 2019-09-16
MAINT: Use isort to add consistency

bashtage:
Use isort to improve import consistency

- [ ] closes #xxxx
- [ ] tests added / passed


addisonlynch:
Can probably update PR template for new formatting requirements (passes ____)

bashtage:
Will alter in a later PR.  Thinking about blackening the code. What do you think?

addisonlynch:
I think that's a good idea...I'm not too familiar with black (other than a little time spent on pandas) but if it takes care of most of the heavy lifting for contributors, all the better. For CI we would use ``black --check``?

bashtage:
Yes, that is right. Black simplifies formatting with contributors.

PR #692, opened 2019-09-18
Rename access_key to api_key

### LINKED ISSUE #620: Inconsistent naming of API key parameter
All ``Reader`` objects have the ``api_key`` attribute which stores their API key (when needed). For example, ``get_data_quandl`` (which uses ``QuandlReader``) accepts ``api_key`` as an argument.

However, one must pass ``access_key`` instead of ``api_key`` to ``DataReader``, which may create confusion ([example](https://stackoverflow.com/questions/54675413/what-is-wrong-with-the-placement-of-my-api-key/54675521#54675521)).

I think it would be best to rename the ``access_key`` for ``DataReader`` to ``api_key`` to create consistent behavior between itself and all other functions in ``data``. 

One other option would be to change the ``api_key`` attribute to ``access_key`` in all other instances, but this is less reasonable in the interest of POLA.

gliptak:

- [x] closes #620 
- [ ] tests added / passed

### DIFF
--- a/docs/source/remote_data.rst
+++ b/docs/source/remote_data.rst
@@ -158,7 +158,7 @@ symbols. The following endpoints are available:
 
     In [4]: f = web.DataReader("AAPL", "av-daily", start=datetime(2017, 2, 9),
        ...:                    end=datetime(2017, 5, 24),
-       ...:                    access_key=os.getenv('ALPHAVANTAGE_API_KEY'))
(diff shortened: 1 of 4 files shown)


addisonlynch:
+1 but we should have a more prominent notif of this in docs. This will be a major breaking change for users who are not using environment variables.

bashtage:
Thanks for catching.  I used a more general method to deprecate it w/o breaking any code.

Issue #694, opened 2019-09-18
RLS: 0.8.0 Release Tracker


bashtage:
A list of issues remaining before a 0.8.0 release:

- [x] Finalize what's new
- [x] Default starting dates #607

bashtage:
@addisonlynch @gliptak Any others?

gliptak:
If ```Python 2,7``` is dropped, documentation and ``setup.py``` would need to be updated accordingly.

addisonlynch:
@bashtage I'd like to do some documentation/readme updates that mention api_key change as well as upcoming end of py2 support

addisonlynch:
I can take care of my updates this evening EST time.

bashtage:
This one is still (and final) 2.7.  

I cut the RC but will happily accept changes.

gliptak:
would you like to make updates date handling https://github.com/pydata/pandas-datareader/pull/677#issuecomment-531997861 ?

Also maybe a further refactoring of Travis https://github.com/pydata/pandas-datareader/pull/668#issuecomment-528640859 ?

addisonlynch:
@gliptak I should be able to resolve those later this evening.

bashtage:
Happy to have Travis improvements.  Although check it first as it has
changed a lot in the past 2 days.

addisonlynch:
@bashtage was there a reason we can't update RTD? I remember you commented something about it in the past, but I can't find it. If not, we should probably remove the reference from the readme.

addisonlynch:
If we can wait until early next week for release, that would be good. I would like to make some changes to reader configuration (default dates, auth) and should have time to do so this weekend.

TL:DR we can (and IMO should) make default dates dynamic & implement as a property. For the base reader we can do something like today - 5 years, and then the other readers can override accordingly. This will allow for arbitrary default start dates for readers which are *not* daily data (such as alphavantage intraday) so if you'd like to do the previous 90 days, that can be implemented.

This should reduce the amount of time we spend updating default dates. Will need to check into the various data sources we have for nuances.

bashtage:
Happy to wait until next week.  5 year rolling is fine with me. Maybe sanitize_dates could have a couple of default start options, e.g, a fixed one, which might make sense for Fama French or a relative one which could take a TimeDelta parsable string (.e.g, '1d'), which would work for intraday.

bashtage:
Will release tomorrow if there are no objections.  Can always do a point release if anything crops up.

[...]

PR #698, opened 2019-09-19
Make requirements more specific

gliptak:

https://github.com/pydata/pandas-datareader/issues/694

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,7 +1,5 @@
-lxml
-pandas>=0.21
-requests>=2.3.0
-wrapt
-ipython
-requests_cache
 sphinx
+ipython
+matplotlib
+sphinx_rtd_theme
+requests_cache
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,9 @@
-ipython
-requests_cache
-sphinx
-sphinx_rtd_theme
-black
-isort
+black; python_version > '3.5'
+flake8-bugbear; python_version > '3.5'
+coverage
+codecov
+coveralls
+flake8
+pytest
+pytest-cov
+wrapt
--- a/ci/pypi-install.sh
+++ b/ci/pypi-install.sh
@@ -1,11 +1,8 @@
 #!/usr/bin/env bash
 
 pip install pip --upgrade
-pip install numpy=="$NUMPY" pytz python-dateutil coverage setuptools html5lib lxml pytest pytest-cov wrapt codecov coveralls beautifulsoup4 isort flake8
-
-if (( $(echo "$TRAVIS_PYTHON_VERSION > 3.5" | bc -l) ));  then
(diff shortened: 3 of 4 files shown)


requirements-doc.txt
```diff
@@ -0,0 +1,6 @@
+sphinx
+ipython
+matplotlib
+sphinx_rtd_theme
+doctr
```
addisonlynch: One small note here...is it necessary to add this to requirements-doc? Contribs don't need to install to work on documentation. It is only needed by Travis. Perhaps we could remove from here and install explicitly in CI

requirements-dev.txt
```diff
@@ -1,6 +1,9 @@
-ipython
-requests_cache
-sphinx
-sphinx_rtd_theme
-black
-isort
+black; python_version > '3.5'
+flake8-bugbear; python_version > '3.5'
```
addisonlynch: We install ``flake8-bugbear`` in CI, but are we using it anywere?

requirements-doc.txt
gliptak: I specifically separated to allow "text" handling of documentation build dependencies.

requirements-dev.txt
gliptak: My understanding is that ```flake8-bugbear``` is a plugin into ```flake8```, so it is a development dependency

bashtage: bugbear is optional and can't be a dependency since it doesn't support Python 2.7.  It can go in after this release.

bashtage: The flake8 'B' tests are bugbear.

gliptak: the specific syntax above should only install it for ```3.5+```

docs/requirements.txt
```diff
@@ -1,7 +0,0 @@
-lxml
```
bashtage: This file has to remain.  It is used by readthedocs.  This should have all docbuild requirement except doctr.

requirements-dev.txt
bashtage: Ignore my previous.  This method is fine.

requirements-doc.txt
bashtage: This file should go into docs/requirements for RTD compat.

bashtage:
One more: doctr needs to be pip installed the if statement

docs/requirements.txt
gliptak: thx. I didn't know RTD uses this. I will update

gliptak:
> 
> 
> One more: doctr needs to be pip installed the if statement

So ```doctr``` needs a minimum Python version?

bashtage:
> So `doctr` needs a minimum Python version?

No, it just isn't a docbuild requirement. It is travis specific, so should only be installed if one travis and in docbuild.

bashtage:
Thanks.

PR #703, opened 2019-09-20
BUG/ENH: Refactor date handling

### LINKED ISSUE #607: IEX Default Start is Invalid
The default start date in the iex "daily.py" module is 1/1/2010, but it is also programmed to give an error if a start date more than five years ago is entered, so if a start date is not provided, the default behavior is to error.  The default start date should be set to be exactly five years ago from the current date, to provide maximum length without error.

addisonlynch:
DNM: need to check docbuild and other readers

PoC re: https://github.com/pydata/pandas-datareader/pull/677#issuecomment-531997861

* Refactors ``_sanitize_dates`` to accept additional date formats (str, int, date, datetime, timestamp) and output a tuple of ``pandas.Timestamp``
* Adds ``default_start_date`` property to allow readers to specify a rolling default start date. ``_BaseReader`` defaults to 15 years. This can be changed (possibly to 5 years).
* Corrects AV and IEX readers to use this paradigm (AV provides "20+ years" so defaults to 20, IEX provides and defaults to 15 years).

- [x] closes #607
- [x] tests added / passed

### DIFF
--- a/docs/source/remote_data.rst
+++ b/docs/source/remote_data.rst
@@ -140,7 +140,9 @@ Historical Time Series Data
 Through the
 `Alpha Vantage <https://www.alphavantage.co/documentation>`__ Time Series
 endpoints, it is possible to obtain historical equities data for individual
-symbols. The following endpoints are available:
+symbols. For daily, weekly, and monthly frequencies, 20+ years of historical data is available. The past 3-5 days of intraday data is also available.
+
+The following endpoints are available:
 
 * ``av-intraday`` - Intraday Time Series
 * ``av-daily`` - Daily Time Series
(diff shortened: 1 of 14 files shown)


addisonlynch:
@bashtage GTG if you and @gliptak are OK with the approach.

gliptak:
@addisonlynch 5 years might be closer to default on many data sources

addisonlynch:
Sounds good. I was leaning that way. I will update.

addisonlynch:
Updated to 5y default. Removed WHATSNEW entry since #607 was introduced after 0.7.0.

gliptak:
Could we consider using ```pd.Datetime``` (sort of a closer match to Pandas)? Larger change ... This thread has some pointers around this:

https://stackoverflow.com/questions/13703720/converting-between-datetime-timestamp-and-datetime64

bashtage:
Do you mean Timestamp?

gliptak:
yes, ```pd.Timestamp``` (sorry for the typo)

bashtage:
Thanks, merging now.  Might be one follow-up.

PR #708, opened 2019-09-22
Update docstring as iso_code is not returned for WorldBank

### LINKED ISSUE #57: pandas.io.wb.download returned DataFrame does not contain iso_code
The [docstring of this method](https://github.com/pydata/pandas-datareader/blob/master/pandas_datareader/wb.py#L112) states that the DataFrame contains the iso_code, which is not the case. The column is [dropped](https://github.com/pydata/pandas-datareader/blob/master/pandas_datareader/wb.py#L159) from the DataFrame before it is returned.

I commented out the above line, but then most values for iso_code are NAN, because of the call to [convert_objects](https://github.com/pydata/pandas-datareader/blob/master/pandas_datareader/wb.py#L166). I think it would be useful to include the iso_code in the result, but if the conversion is to be kept, it would need to become part of the index.

Not sure, what would be the best solution here. In case the current behaviour is to be kept, the docstring should be corrected.

I previously reported this here: https://github.com/pydata/pandas/issues/10860

gliptak:
- [x] closes #57

```
from pandas_datareader import wb
>>> pandas_datareader.__version__
'0.8.0'
>>> ind = ['NY.GDP.PCAP.KD', 'IT.MOB.COV.ZS']
>>> dat = wb.download(indicator=ind, country='all', start=2015, end=2015).dropna()
/home/jovyan/.local/lib/python3.7/site-packages/pandas_datareader/wb.py:659: UserWarning: No results found from world bank. Indicator: IT.MOB.COV.ZS
  warnings.warn(msg)
>>> dat
                                     NY.GDP.PCAP.KD
country                        year                
Arab World                     2015     6430.464609
Caribbean small states         2015     9224.146707
Central Europe and the Baltics 2015    14327.577261
Early-demographic dividend     2015     3400.049930
East Asia & Pacific            2015     9227.673906
...                                             ...
Virgin Islands (U.S.)          2015    29408.055125
West Bank and Gaza             2015     2631.811985
Yemen, Rep.                    2015      908.757846
Zambia                         2015     1641.005482
Zimbabwe                       2015     1234.103352

[244 rows x 1 columns]
```

### DIFF
--- a/pandas_datareader/wb.py
+++ b/pandas_datareader/wb.py
@@ -856,7 +856,7 @@ def download(
     Returns
     -------
     data : DataFrame
-        DataFrame with columns country, iso_code, year, indicator value
+        DataFrame with columns country, year, indicator value
     """
     return WorldBankReader(
         symbols=indicator,


bashtage:
Thanks.

PR #718, opened 2019-10-01
fix for futures

### LINKED ISSUE #717: stooq.com - Problem getting data when symbol name ends with ".*"
First, thanks for the great module!

When getting data for symbol that ends with .* (CLX19.F)  `.US` is added to symbol, since `f`
is not present in the `["de", "hk", "hu", "jp", "uk", "us"]` array

This works (ohlc timeseries is returned):
`curl "https://stooq.com/q/d/l/?s=CLX19.F&d1=20190101&d2=20190930&i=d"`

This returns `No data`:
`curl "https://stooq.com/q/d/l/?s=CLX19.F.US&d1=20190101&d2=20190930&i=d"`

Workaround would be to add `"f"` in the array where check is being made

https://github.com/pydata/pandas-datareader/blob/48a538c14041bd166a4d138dd3ee372e76f6d2ab/pandas_datareader/stooq.py#L45-L51

I can provide PR with aforementioned workaround.

kdcro101:
closes #717

### DIFF
--- a/pandas_datareader/stooq.py
+++ b/pandas_datareader/stooq.py
@@ -48,7 +48,13 @@ class StooqDailyReader(_DailyBaseReader):
                 symbol = symbol_parts[0]
             else:
                 if symbol_parts[1].lower() not in [
-                    "de", "hk", "hu", "jp", "uk", "us", "f"
+                    "de",
+                    "hk",
+                    "hu",
+                    "jp",
+                    "uk",
+                    "us",
+                    "f",
                 ]:
                     symbol = ".".join([symbol, "US"])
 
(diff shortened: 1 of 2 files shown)


bashtage:
Can you add a test that reads a symbol that causes the problem, and a note to the what's new? (THere might not be a what's new for 0.9 yet, in which case you can skip the note).

kdcro101:
@bashtage ,

I've added test for `CLX19.F`.

Can you help me with travis-ci failures?

kdcro101:
@bashtage,

I've checked... The coverage is unchanged for files that I've changed...

bashtage:
Thanks.

PR #729, opened 2019-11-07
Stooq data dates closes #578

### LINKED ISSUE #578: Stooq: 1) Data before April 2010 is not shown 2) start / end does not work
I have two problems with version 0.7.0 and Data from Stooq:

1) Data before April 2010 ist not shown (works in version 0.6.0)
2) start and end does not work (same problem in version 0.6.0)

Here is the code:
`import pandas as pd
symbol ='AAPL.US'
start = '20000101'
end   = '20180901'
df = web.DataReader(symbol, 'stooq', start, end)
print(df)`

simongarisch:
closes #578


bashtage:
Thanks.

PR #737, opened 2019-12-11
update EcondbReader.read to use session

### LINKED ISSUE #736: econDB ignores session info
When using econDB it ignores session info passed to it.
I believe econdb.py line 26:
`results = requests.get(self.url).json()["results"]`
should be:
`results = self.session.get(self.url).json()["results"]`
At least that makes it work for me

grwoodw:
- [ ] closes #736
- [ ] tests added / passed


bashtage:
Failure looks unrelated. Thanks.

PR #745, opened 2020-01-20
Added better error logging for API limit threshold

PatrickAlphaVantage:
- [ ] closes #xxxx
- [ ] tests added / passed

Now when the API rate limit is hit instead of the error being:
```
--> 81                 raise RemoteDataError()
     82         df = df[sorted(df.columns)]
     83         df.columns = [id[3:] for id in df.columns]

RemoteDataError: 
```

It will be:
```
pandas_datareader._utils.RemoteDataError: {'Note': 'Thank you for using Alpha Vantage! Our standard API call frequency is 5 calls per minute and 500 calls per day. Please visit https://www.alphavantage.co/premium/ if you would like to target a higher API call frequency.'}
```

### DIFF
--- a/pandas_datareader/av/__init__.py
+++ b/pandas_datareader/av/__init__.py
@@ -78,7 +78,10 @@ class AlphaVantage(_BaseReader):
                     ".".format(self.symbols)
                 )
             else:
-                raise RemoteDataError(out)
+                raise RemoteDataError(
+                    " Their was an issue from the data vendor "
+                    "side, here is their response: {}".format(out)
+                )
         df = df[sorted(df.columns)]
         df.columns = [id[3:] for id in df.columns]
         return df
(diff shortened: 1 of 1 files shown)


addisonlynch:
@PatrickAlphaVantage thanks!

PatrickAlphaVantage:
So it looks like the builds were failing before I made this PR, is master broken at the moment?  Or did I do something silly.

addisonlynch:
@PatrickAlphaVantage yeah failing for different reasons

pandas_datareader/av/__init__.py
```diff
@@ -78,7 +78,7 @@ def _read_lines(self, out):
                     ".".format(self.symbols)
                 )
             else:
-                raise RemoteDataError()
+                raise RemoteDataError(out)
```
bashtage: Rather than just dump out, it would be better to explain that this is the content retrieved from the website and then show out. Basically a formatted string is needed around out.

PatrickAlphaVantage: Good call, fixed!

bashtage:
Thanks.

PR #749, opened 2020-01-30
Fama French data: Update regex search for finding table start to exclude comment strings

jayvn:
Fixes crash when reading Fama-French data (US Momentum data  specifically)


bashtage:
Was this a temporary issue?

PR #790, opened 2020-06-18
add timeout

### LINKED ISSUE #789: Timeout not actually used
I noticed my downloader hung forever. I dug through the code and it looks like the timeout parameter isn't actually passed the request call.

https://github.com/pydata/pandas-datareader/blob/09ab475788daee235e0a7d238607cf674e1c19c7/pandas_datareader/base.py#L142

Pretty sure the fix is just:

```
- response = self.session.get(url, params=params, headers=headers)
+ response = self.session.get(url, params=params, headers=headers, timeout=self.timeout)
```

JesseAldridge:
- [x] closes #789
- [ ] tests added / passed

### DIFF
--- a/pandas_datareader/base.py
+++ b/pandas_datareader/base.py
@@ -152,7 +152,9 @@ class _BaseReader(object):
         pause = self.pause
         last_response_text = ""
         for _ in range(self.retry_count + 1):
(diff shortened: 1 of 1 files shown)


bashtage:
Thanks.

Issue #748, opened 2020-01-26
MOEX: open, high, low - incorrect


Curious75:
ticker = "MAGN" 2014-12-30
datareader extract other values than we can get from other resources

for example you can compare to MOEX-site: 
http://www.moex.com/ru/issue.aspx?board=TQBR&code=MAGN
LOW = Цена сделки минимальная | 10,061 (datareader = 8,8384)
HIGH = Цена сделки максимальная | 10,947 (datareader = 12,1528)
OPEN = Цена предторгового периода/Цена аукциона открытия | 10,947  (datareader = 12,1528)

CLOSE the same = Цена закрытия | 10,821 (datareader = 10,821)

you can also compare with https://www.finam.ru/profile/moex-akcii/mmk/export/?

GreenArt11:
`pdr.DataReader('MAGN', 'moex', '2014-12-30', '2014-12-30')` returns 13 rows for the results in every of the 13 boards and trading modes.

You should use only rows with the link to `TQBR` board due to it is the main board of this ticker (as it is shown there: `https://iss.moex.com/iss/securities/MGNT`)

Made an improvement issue #811

GreenArt11:
fixed by #812

PR #812, opened 2020-07-21
MOEX: Use data from the main board only for each ticker

### LINKED ISSUE #811: MOEX: Replace data from all boards to the data from the main board
Loading data for Moscow exchange returns info from all of the boards and trading modes. 
```
pdr.DataReader('sber', 'moex', '2020-07-02', '2020-07-02')
```
The code returns 9 rows for each board (trading mode) for the same date. 
It seems to be better if pdr could return data from the main board only (for each ticker). It would lead to less misunderstanding the results.

I also would like to try this as my first PR

GreenArt11:
- [X] closes #811 and #748
- [X] tests added / passed
- [X] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [X] passes `black --check pandas_datareader`

Filters output using only data from the main board for every ticker (before that it was returning unuseful stats from repo boards and null values from boards, that were not main for the asset and had no trading history).

### DIFF
--- a/docs/source/remote_data.rst
+++ b/docs/source/remote_data.rst
@@ -680,11 +680,19 @@ MOEX Data
 =========
 The Moscow Exchange (MOEX) provides historical data.
 
+pandas_datareader.get_data_moex(*args) is equivalent to 
+pandas_datareader.moex.MoexReader(*args).read() 
 
 .. ipython:: python
 
-   import pandas_datareader.data as web
-   f = web.DataReader('USD000UTSTOM', 'moex', start='2017-07-01', end='2017-07-31')
+   import pandas_datareader as pdr
+   f = pdr.get_data_moex(['USD000UTSTOM', 'MAGN'], '2020-07-02', '2020-07-07')
+   f.head()
+
+   f = pdr.moex.MoexReader('SBER', '2020-07-02', '2020-07-03').read()
+   f.head()
+   
+   f = pdr.moex.MoexReader('SBER', '2020-07-02', '2020-07-03').read_all_boards()
    f.head()
 
 .. _remote_data.naver:
--- a/docs/source/whatsnew.rst
+++ b/docs/source/whatsnew.rst
@@ -18,6 +18,7 @@ What's New
 
 These are new features and improvements of note in each release.
 
+.. include:: whatsnew/v0.9.1.txt
 .. include:: whatsnew/v0.9.0.txt
 .. include:: whatsnew/v0.8.0.txt
 .. include:: whatsnew/v0.7.0.txt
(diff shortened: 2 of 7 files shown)


GreenArt11:
If we try to load data for a single date and ticker:
```
data = pdr.data.DataReader('Sber', 'moex', '2020-07-02', '2020-07-02')
data[['BOARDID', 'SECID', 'NUMTRADES', 'CLOSE', 'VOLUME']]
```
we get this output:
```
BOARDID | SECID | NUMTRADES | CLOSE | VOLUME
EQRP | SBER | 91.0 | 4.25 | NaN
PSRD | SBER | 2.0 | 0.65 | NaN
PSRP | SBER | 136.0 | 4.50 | NaN
PTEQ | SBER | 8.0 | 210.00 | 914392.0
RPMO | SBER | 951.0 | 209.50 | NaN
SOTC | SBER | 15.0 | NaN | NaN
MXBD | SBER | NaN | NaN | NaN
SMAL | SBER | 20.0 | 209.90 | 49.0
TQBR | SBER | 78824.0 | 209.60 | 76382050.0
```
Most of the boards return NaNs or a little amount of trades. This is night trades or trades for qualified investors. Moreover, some of the non-primary boards are traded in other currency than the primary for the ticker.
This PR makes output look like this:
```
BOARDID | SECID | NUMTRADES | CLOSE | VOLUME
TQBR | SBER | 78824 | 209.6 | 76382050
```
We get a single row for one ticker for one date - only from the primary board. Others are excluded.

There could be a chance that a person is willing to get all data from the boards, although I have not met this cases. On the contrary, a lot of posts on finance forums for amateur investors are about this case like described in the issue #748.

GreenArt11:
> Will need a release note since this a breaking change. Is there any reason to want the other boards? Do they ever differ? If so, could retrieving them all be an option?

I can also try to make a specific def for the old variant of the output for possible backward compatibility with the module. It would still return only one-per-day-per-ticker data in a default query. But it would also be possible to make a specific query for the result as it returns now.

GreenArt11:
Added function `read_all_boards()` for the class in Moex module that provides data from all of the boards (as it is now for `read()` in v.0.9.0). 
Default behaviour for `read()` function changed to loading data from the primary board only.

GreenArt11:
@bashtage Hi! Can you give me some advice how to increase coverage to pass the coveralls test?
I've change only two files and coverage slightly changed, although coveralls reports `318 existing lines in 13 files now uncovered`. I am not sure how to solve this.

bashtage:
We never pass that :-<

bashtage:
Looks good. Could you do the what's new and then once Travis passes it will be in.

GreenArt11:
Thank you! Should I start new docs/source/whatsnew/v0.10dev.0.txt file?

bashtage:
Can you add it to 0.9.1. I think I'll get it out before 0.10.

GreenArt11:
Created v.0.9.1 What's new and also made some docs improvements. All of them are described in the What's new file.

bashtage:
Thanks!

PR #814, opened 2020-08-02
Fixing tests for index.freq

simongarisch:
Fixing tests.

pandas.testing.assert_series_equal now includes a default argument of [check_freq=True](https://github.com/pandas-dev/pandas/blob/master/pandas/_testing.py#L1217). With this new arg some of the tests are failing as the index.freq does not match in comparisons.

I've adjusted index.freq so that the tests pass. Another option is to use check_freq=False, but this argument is not available in earlier versions of pandas. Alternatively, we could inspect the function for the existence of check_freq and pass False if required. I considered changing index.freq to be the easiest option.

Thanks


bashtage:
Looks good to me.  Thanks!

PR #818, opened 2020-08-29
Fixed test_moex_stock_datareader

simongarisch:
It looks like extra columns are coming back from the api and causing the df.size assertion to fail. I've changed this to assert the number of rows == 2 for each of "GAZP", "SIBN".

Cheers


bashtage:
Thanks.

PR #819, opened 2020-08-30
Create common.py

aarjaneiro:
- [ ] closes #xxxx
- [ ] tests added / passed

Currently, packages like quantmod are broken without this. I think it would be wise to have either common readded or urlencode readded or at least as an optional dependency.

### DIFF
--- a/pandas_datareader/io/common.py
+++ b/pandas_datareader/io/common.py
@@ -1,0 +1,550 @@
+"""Common IO api utilities"""
+
+import os
+import csv
+import codecs
+import mmap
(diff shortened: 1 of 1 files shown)


bashtage:
Why is this needed?

bashtage:
If quantmod is broken then this fix should go into quantmod.

aarjaneiro:
> If quantmod is broken then this fix should go into quantmod.

Fair enough. I am unsure as to why common was deleted or how many other packages might have been effected, but the burden for alleviating dependencies indeed does not lay here.

PR #826, opened 2020-09-10
Econdb and Enigma mismatch on bookmarks fixed

githubsaturn:
- [x ] added entry to docs/source/whatsnew/vLATEST.txt


econdb:
thanks @bashtage

PR #850, opened 2021-02-16
Fix the unstacking when data contains duplicated rows with the same index

### LINKED ISSUE #738: double dates for some symbols from yahoo is leading to unstacking error.
basically the issue is described here:

https://stackoverflow.com/questions/57885790/valueerror-index-contains-duplicate-entries-cannot-reshape-pandas-datareader/59347328#59347328

I tried to download SP500 from yahoo. Following symbols (CBS, STI, VIAB) led to an unstacking error in pandas-datareader. To reproduce error download CBS and any other Symbol together for Dec 2019 as the error happen for Dec data (in case of CBS we have two values from yahoo for the 4th Dec).

Unstacking error doesn't appear when CBS is downloaded by its own.

Regards

changlan:
- [x] closes #738
- [x] tests added / passed
- [x] added entry to docs/source/whatsnew/v0.9.1.txt

### DIFF
--- a/docs/source/whatsnew/v0.9.1.txt
+++ b/docs/source/whatsnew/v0.9.1.txt
@@ -20,13 +20,15 @@ Enhancements
   from every trading board (ver. 0.9.0 behaviour)
 - Docs for MOEX reedited
 
-.. _whatsnew_080.bug_fixes:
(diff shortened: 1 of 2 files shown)


scribu:
Can confirm that this fixes the issue 👍

Issue #610, opened 2019-01-26
[BUG] Yahoo duplicate row


okomarov:
I am on 0.7.0 pdr:

```
pdr.get_data_yahoo('AMZN', start='2019-01-25', end='2019-01-26')
```
where `2019-01-26` is a Saturday.

This outputs:
```
                   High          Low    Open        Close   Volume    Adj Close
Date
2019-01-25  1683.449951  1662.000000  1670.5  1670.569946  4959679  1670.569946
2019-01-25  1683.479980  1661.609985  1670.5  1670.569946  4945900  1670.569946
```

It should only return the last row.

okomarov:
This seems to be an issue with yahoo itself.

Querying same day, returns nothing:
<br>

<img width="934" alt="screenshot 2019-01-26 at 22 04 51" src="https://user-images.githubusercontent.com/3731173/51792843-3729f200-21af-11e9-8c62-4fd0b9ad282f.png">

Querying Friday to Sunday returns the two rows listed above:
<br>
<img width="943" alt="screenshot 2019-01-26 at 21 56 31" src="https://user-images.githubusercontent.com/3731173/51792861-70626200-21af-11e9-9008-5415cf96d575.png">

and params for the GET request are:

 

  name | value 
-- | --
formatted | true
crumb | plThqsOcXUi
lang | en-US
region | US
period1 | 1548374400
period2 | 1548547200
interval | 1d
events | div\|split
corsDomain | finance.yahoo.com

okomarov:
Reported bug on [Yahoo forums](https://yahoo.uservoice.com/forums/382977-finance/suggestions/36651802-duplicate-row)

bashtage:
Closing as answered.

PR #858, opened 2021-04-03
Fix duplicated rows

### LINKED ISSUE #738: double dates for some symbols from yahoo is leading to unstacking error.

dangillet:
Duplicate of PR #850 fixing the line length issue for a comment

- [ ] closes #738 and #610
- [ ] tests added / passed

### DIFF
--- a/docs/source/whatsnew/v0.9.1.txt
+++ b/docs/source/whatsnew/v0.9.1.txt
@@ -20,13 +20,15 @@ Enhancements
   from every trading board (ver. 0.9.0 behaviour)
 - Docs for MOEX reedited
 
-.. _whatsnew_080.bug_fixes:
+.. _whatsnew_091.bug_fixes:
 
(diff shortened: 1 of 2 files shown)


dangillet:
I realise now that travis fails for other reason. This PR is muddying the water. I'll close it and I'll leave it to the original PR.

Issue #867, opened 2021-07-02
_get_response without headers doesn't work (at least with 'yahoo' source


galashour:
to fix, I put in base.py:

    def _get_response(self, url, params=None, headers=None):
        """ send raw HTTP request to get requests.Response from the specified url
        Parameters
        ----------
        url : str
            target URL
        params : dict or None
            parameters passed to the URL
        """

        # initial attempt + retry
        if headers == None:
            headers          = {'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/39.0.2171.95 Safari/537.36'}

        pause = self.pause

ivofernandes:
Thanks, I tested here, it works like that, I was also with an error since yesterday

maxmekiska:
Hi everyone, I am getting an error as well when I try to retrieve stock data from yahoo finance.  Is this error related to the same issue as described above?

```
RemoteDataError: Unable to read URL: https://finance.yahoo.com/quote/F/history?period1=1264996800&period2=1517543999&interval=1d&frequency=1d&filter=history
Response Text:
[...]

PR #869, opened 2021-07-05
Update daily.py

s-meinhardt:
**Why:**
Yahoo Finance must have changed their API in the last few weeks causing errors or no returned data.

**What:**
I've changed the default interval to `"1d"` instead of `"d"` and added the missing headers in the `_get_response` call in the `_read_one_data` method of the `YahooDailyReader` class. With the proposed change, I was able to get the data from Yahoo's Finance API. Note that other intervals are effected too, but the DataReader class seems to use `"d"` only.

### DIFF
--- a/pandas_datareader/yahoo/daily.py
+++ b/pandas_datareader/yahoo/daily.py
@@ -63,7 +63,7 @@ class YahooDailyReader(_DailyBaseReader):
         adjust_price=False,
         ret_index=False,
         chunksize=1,
-        interval="d",
+        interval="1d",
         get_actions=False,
         adjust_dividends=True,
     ):
--- a/pandas_datareader/yahoo/daily.py
+++ b/pandas_datareader/yahoo/daily.py
@@ -150,7 +150,7 @@ class YahooDailyReader(_DailyBaseReader):
         del params["symbol"]
         url = url.format(symbol)
 
-        resp = self._get_response(url, params=params)
+        resp = self._get_response(url, params=params, headers=self.headers)
         ptrn = r"root\.App\.main = (.*?);\n}\(this\)\);"
         try:
             j = json.loads(re.search(ptrn, resp.text, re.DOTALL).group(1))


bashtage:
Does this fix work locally?

s-meinhardt:
I also had to add the headers to the `_get_response` call in the `_read_one_data` function which I didn't in the first commit. The missing headers raised the `RemoteDataError` in the `_get_response` function. Once the headers are added, the interval update is necessary to get a non-empty DataFrame. After these two changes, it was running fine locally and returned the price data etc.

ralexx:
Merging 389ce32 will break the use case of passing a correctly parameterized `requests.Session` instance to `YahooDailyReader` via the `session` argument. Second, the change will further encourage the user anti-pattern of relying on a hard-coded user agent string, with a high likelihood of further issues being opened if said user agent string is blocked. Third, should the API owners implement robot blocking by filtering user agent strings for statistical outliers, hard-coding an arbitrary browser string will quickly make that string an outlier as its browser version ages.

All of these drawbacks can be avoided simply by pointing users to a documented [recipe](https://github.com/pydata/pandas-datareader/issues/867#issuecomment-873215586) to pass in their choice of user agent string through the existing `YahooDailyReader.session` attribute as described above. `requests` is already present where `pandas-datareader` is installed, so no additional dependencies are created. I'm willing to open a PR for the doc change.

### Problem example

When the `headers` parameter is passed to `requests.Session.get`, it replaces any headers that have already been bound to the `Session` instance. Anyone using a parameterized `Session` as I do will be forced to monkey patch their headers to `YahooDailyReader.headers` if 389ce32 is merged.

```python
# from pandas_datareader.yahoo.daily import YahooDailyReader 
# import requests

USER_AGENT = {
    'User-Agent': ('Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko)'
                   ' Chrome/91.0.4472.124 Safari/537.36')
    }
my_headers = {**various_headers, **USER_AGENT}
sesh = requests.Session()
sesh.proxy = {<proxy settings>}  # one of the reasons I need to pass a Session instance to YDR
sesh.headers = my_headers


reader = YahooDailyReader(session=sesh, **other_kwargs)
# now I will have to monkey patch my headers
reader.headers = my_headers
df = reader.read()
```

Even if you disagree with me and prefer to enable the use of a 'fallback' hard-coded user agent string, please do not implement it as proposed in 389ce32. At a bare minimum, add a `headers=None` argument to `YahooDailyReader.__init__` to remain consistent with the existing class pattern where all instance attributes except `pause_multiplier` are set via init arguments. Then move the default header to be a class attribute and bind it to the instance only where the default `headers=None` is passed.

s-meinhardt:
Hi @radexx, 

I understand your point and see why one might want to have a user defined header. I'll close this PR and continue to contribute at [PR 873](https://github.com/pydata/pandas-datareader/pull/873) attacking the same issue.

Issue #868, opened 2021-07-03
data_source='yahoo': reading data fails since July 01


aeolio:
Since 2021-07-01, reading data from Yahoo fails with a nondescript error message. 

Calling the URL (see RemoteDataError below) directly from a browser succeeds. wget returns error 404: Not Found.

Interesting is the name of the background image in the HTML message: "s.yimg.com/nn/img/**sad-panda**-201402200631.png"

_Pandas datareader version 0.9.0_

Code snippet:
```
import pandas as pd
import pandas_datareader as pdr
# start date is arbitrary
pdr.DataReader('GOOGL', data_source='yahoo', start='2021-07-01')
```
Error message:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/iago/.local/lib/python3.9/site-packages/pandas/util/_decorators.py", line 199, in wrapper
    return func(*args, **kwargs)
  File "/usr/lib/python3.9/site-packages/pandas_datareader/data.py", line 376, in DataReader
    return YahooDailyReader(
  File "/usr/lib/python3.9/site-packages/pandas_datareader/base.py", line 253, in read
    df = self._read_one_data(self.url, params=self._get_params(self.symbols))
  File "/usr/lib/python3.9/site-packages/pandas_datareader/yahoo/daily.py", line 153, in _read_one_data
    resp = self._get_response(url, params=params)
  File "/usr/lib/python3.9/site-packages/pandas_datareader/base.py", line 181, in _get_response
    raise RemoteDataError(msg)
pandas_datareader._utils.RemoteDataError: Unable to read URL: https://finance.yahoo.com/quote/GOOGL/history?period1=1625104800&period2=1625363999&interval=1d&frequency=1d&filter=history
Response Text:
[...]

PR #873, opened 2021-07-10
resolve issue #867 by explicitly adding headers to url _get_response

galashour:
- [ ] closes #xxxx
- [ ] tests added / passed

### DIFF
--- a/pandas_datareader/base.py
+++ b/pandas_datareader/base.py
@@ -47,15 +47,15 @@ class _BaseReader(object):
     _format = "string"
 
     def __init__(
-        self,
-        symbols,
-        start=None,
-        end=None,
-        retry_count=3,
-        pause=0.1,
-        timeout=30,
-        session=None,
-        freq=None,
+            self,
+            symbols,
+            start=None,
+            end=None,
+            retry_count=3,
+            pause=0.1,
+            timeout=30,
+            session=None,
+            freq=None,
     ):
 
         self.symbols = symbols
@@ -149,6 +149,9 @@ class _BaseReader(object):
         """
 
         # initial attempt + retry
+        if headers == None:
+            headers = self.headers
+
         pause = self.pause
         last_response_text = ""
         for _ in range(self.retry_count + 1):
@@ -224,14 +227,14 @@ class _DailyBaseReader(_BaseReader):
     """ Base class for Google / Yahoo daily reader """
 
     def __init__(
-        self,
-        symbols=None,
-        start=None,
-        end=None,
-        retry_count=3,
-        pause=0.1,
-        session=None,
-        chunksize=25,
+            self,
+            symbols=None,
+            start=None,
+            end=None,
+            retry_count=3,
+            pause=0.1,
+            session=None,
+            chunksize=25,
     ):
         super(_DailyBaseReader, self).__init__(
             symbols=symbols,
@@ -297,7 +300,7 @@ def _in_chunks(seq, size):
     """
     Return sequence in 'chunks' of size defined by size
     """
-    return (seq[pos : pos + size] for pos in range(0, len(seq), size))
+    return (seq[pos: pos + size] for pos in range(0, len(seq), size))
 
 
 class _OptionBaseReader(_BaseReader):
@@ -331,7 +334,7 @@ class _OptionBaseReader(_BaseReader):
         raise NotImplementedError
 
     def get_near_stock_price(
-        self, above_below=2, call=True, put=False, month=None, year=None, expiry=None
+            self, above_below=2, call=True, put=False, month=None, year=None, expiry=None
     ):
         """
         ***Experimental***
@@ -340,7 +343,7 @@ class _OptionBaseReader(_BaseReader):
         raise NotImplementedError
 
     def get_forward_data(
-        self, months, call=True, put=False, near=False, above_below=2
+            self, months, call=True, put=False, near=False, above_below=2
     ):  # pragma: no cover
         """
         ***Experimental***
--- a/pandas_datareader/yahoo/daily.py
+++ b/pandas_datareader/yahoo/daily.py
@@ -150,7 +150,7 @@ class YahooDailyReader(_DailyBaseReader):
         del params["symbol"]
         url = url.format(symbol)
 
-        resp = self._get_response(url, params=params)
+        resp = self._get_response(url, params=params, headers=self.headers)
         ptrn = r"root\.App\.main = (.*?);\n}\(this\)\);"
         try:
             j = json.loads(re.search(ptrn, resp.text, re.DOTALL).group(1))


galashour:
I'm relatively new, the fix is simple and addresses issue #867 , without the fix, using yahoo source fails (/crashes).

I'm not sure if / how to do some of the above, but I did test that before the fix the library crashes and after the fix it completes. Also others commenting on issue #867 confrimed the same.

I'm not sure if/how to performm the above checkboxes, and in which environment, possibly you guys can help.

s-meinhardt:
Hi galashour,

I made a similar PR a couple of days ago [here](https://github.com/pydata/pandas-datareader/pull/869) but will close mine as your solution seems to tackle the problem described by @ralexx in my PR. 

However, you should test if you actually get any data back and not just an empty DataFrame. I've noticed that I had to change the default interval from `'d'` to `'1d'` in the `daily.py` file too in order to get actual data. See my commit [#869](https://github.com/pydata/pandas-datareader/pull/869) for further details. You might want to add another commit to fix that issue in your PR.

galashour:
> Hi galashour,
> 
> I made a similar PR a couple of days ago [here](https://github.com/pydata/pandas-datareader/pull/869) but was made aware of an issue by @ralexx with whom you had a similar discussion [here](https://github.com/pydata/pandas-datareader/issues/867#issuecomment-873238509). As there is no reason to have multiple PRs for the same problem, I'll close mine and continue to contribute here. I hope the issue raised by @ralexx can be solved soon as I want the YDR back to work.
> 
> However, you should test if you actually get any data back and not just an empty DataFrame. I've noticed that I had to change the default interval from `'d'` to `'1d'` in the `daily.py` file too in order to get actual data. See my commit [#869](https://github.com/pydata/pandas-datareader/pull/869) for further details. You might want to add another commit to fix that issue in your PR.

Thanks for the comment. 
I verified that I do get valid daily data once the header issue was addressed.
>> df1 = web.DataReader(ticker1, data_source='yahoo', start=date_start, end=date_end)

s-meinhardt:
I can also get data now without the interval modification. Yahoo must have fixed their API as it definitely didn't work at the beginning of the week. Well, I hope you get your PR accepted asap.

ralexx:
Please do not merge this pull request as it appears now. It will break the idiomatic use of user-defined `requests.Session` instances in pandas_datareader. A much simpler change can achieve the desired result without breaking any use cases.

I appreciate that @galashour is intent on embedding a default user agent string in pandas_datareader. To be clear, I have already [submitted](https://github.com/pydata/pandas-datareader/issues/867#issuecomment-873215586) that this is both unnecessary and an anti-pattern. But leaving aside my protest, this patch currently takes a harmful approach that will unnecessarily break valid use cases.

### Problem

These changes force all users to monkey patch `pandas_datareader.yahoo.daily.YahooDailyReader.headers` if they do not want to use the request headers arbitrarily hard-coded into that module:

```python
        self.headers = {
            "Connection": "keep-alive",
            "Expires": str(-1),
            "Upgrade-Insecure-Requests": str(1),
            # Google Chrome:
            "User-Agent": (
                "Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 "
                "(KHTML, like Gecko) Chrome/54.0.2840.99 Safari/537.36"
            ),
}
```

This is because in `requests` [my emphasis in the quote that follows],

> Any dictionaries that you pass to a request method will be merged with the session-level values that are set. *The method-level parameters override session parameters*. [[docs]](https://docs.python-requests.org/en/master/user/advanced/#session-objects)

The proposed change to `YahooDailyReader._read_one_data` causes the `headers` argument to be passed to the `pandas_datareader.base._BaseReader.session.get` *method*, which overrides any headers that may have been set on `_BaseReader.session`. That includes the headers on any `requests.Session` instance passed to `YahooDailyReader`'s `session` argument (which is passed up to `_BaseReader.session` as the intervening subclasses call `super()`).

Because no public `headers` argument exists in `YahooDailyReader.__init__`, a monkey patch would be required to avoid being compelled to use the hard-coded headers. Even if you add a public API for setting the headers, users will still be forced to pass their preferred headers to each `YahooDailyReader` instance. This defeats the design intent of a `requests.Session`: "Sessions can also be used to provide default data to the request methods." [[docs]](https://docs.python-requests.org/en/master/user/advanced/#session-objects)

### Non-breaking approach

If you remain intent on providing a hard-coded, fallback user agent string, it can be accomplished without breaking any use case.

Note that the `_BaseReader.session` attribute is set by calling `pandas_datareader._utils._init_session`. If no `Session` instance has been passed to `_BaseReader.__init__`, `_init_session` instantiates one and binds `_BaseReader.session` to it. Make changes here and they will only be applied where users do not pass a `Session` instance.

```python
# pandas_datareader._utils.py

DEFAULT_USER_AGENT = <some user agent string>  # so far this is the only change needed for Yahoo API

def _init_session(session, retry_count=3):
    if session is None:
        session = requests.Session()
        # do not set requests max_retries here to support arbitrary pause
        session.headers['User-Agent'] = DEFAULT_USER_AGENT  # <-- NON-BREAKING CHANGE
    return session
```

galashour:
There is No HARD CODED header content in the fix, not sure if you actually read the actual fix.

It puts on line 153 of the yahoo/daily.py the 3rd parameter that was missing, with NO hardcoded string:
resp = self._get_response(url, params=params, headers=self.headers)

Similarly, it also put a safeguard in the base class (again happens to be line 153) only for the case somehow we end up with headers == None:
        if headers == None:
            headers = self.headers

ralexx:
> There is No HARD CODED header content in the fix

`self.headers` in your patch is `pandas_datareader.yahoo.daily.YahooDailyReader.headers` (lines 84-93). There is no interface to set this value.

As I said, even if you sort this issue out, your patch still forces all users to pass headers if they don't want to use the headers as defined above. There are other ways of accomplishing your purpose without disadvantaging users.

galashour:
once again - wrong. Users don't provide headers, and I didn't provide any headers in my testing, and no interface was changes whatsoever.

ralexx:
Please be kind, I am trying to help you. I understand that your changes work for you, but so far you have not provided any tests for your proposed changes so it's not clear that your changes work for other use cases. As the author of a PR it's incumbent upon you to address fellow users' concerns.

You propose a change that violates the public API of `pandas_datareader.yahoo.daily.YahooDailyReader` and I have explained why this is the case. If you believe I am wrong then your tests should show that clearly. Just pass a `requests.Session` instance into `YahooDailyReader` and show that `YahooDailyReader(**kwds).read()` fails (because we know that Yahoo is rejecting the default user agent string 'User-Agent': 'python-requests/2.25.1' used by `requests.Session`.

> the interface in the official build is broken when working with yahoo

This is not accurate. I demonstrated this to you in #867. Anyone using version 0.9 can download historical prices today by passing a `requests.Session` instance *with a valid user agent header* into the `session` argument of the `YahooDailyReader.__init__ interface`.

Or, by making a trivial method call, you can even use the same hard-coded request headers on lines 84-93 of yahoo/daily.py that your patch points to:

```python
from pandas_datareader.yahoo.daily import YahooDailyReader

reader = YahooDailyReader('MSFT', start='2021-07-01', end='2021-07-09')

# confirm that the default headers are the ones we don't want
print(reader.session.headers)

# {'User-Agent': 'python-requests/2.25.1', 'Accept-Encoding': 'gzip, deflate', 'Accept': '*/*', 'Connection': 'keep-alive'}

# now update the *session* headers from `YahooDailyReader.headers`
reader.session.headers.update(reader.headers)
print(reader.session.headers)

# {'Connection': 'keep-alive',
#  'Expires': '-1',
#  'Upgrade-Insecure-Requests': '1',
#  'User-Agent': 'Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/54.0.2840.99 Safari/537.36'}

# download prices as usual
df = reader.read()
print(df)

#                  High         Low        Open       Close    Volume   Adj Close
# Date
# 2021-07-01  271.839996  269.600006  269.609985  271.600006  16725300  271.600006
# 2021-07-02  278.000000  272.500000  272.820007  277.649994  26458000  277.649994
# 2021-07-06  279.369995  274.299988  278.029999  277.660004  31565600  277.660004
# 2021-07-07  280.690002  277.149994  279.399994  279.929993  23260000  279.929993
# 2021-07-08  278.730011  274.869995  276.899994  277.420013  24618600  277.420013
# 2021-07-09  278.049988  275.320007  275.720001  277.940002  23905500  277.940002
```

Please confirm that you are able to run this code.

galashour:
The API that used to work was simple and is used by many:
>> df1 = web.DataReader(ticker1, data_source='yahoo', start=date_start, end=date_end)

This API is now broken.
It is a common API used, by many.

It is not only for me, quite a few on issue #867 (and also on issue #868 that is a duplicate) - reported the same problem, and reported it works for them with the simple fix applied.

Original API was broken due to a 'bug' in the existing that yahoo.daily as I mentioned that 'forgot' to add the self.headers that was already set internally correctly (with no explicit value from the users), while a preferred practice regrdless to yahoo is to use headers in the url-request (as can be seen also in the base class).

I will say it one last time, as nicely as I can, I didn't violate any public API (strong and unjustified claims), nor did I change any API (public or internal). Simply added a parameter that was already there unused while it should have been used.
Also the base class does use the headers in the url-request, and just the yahoo-daily implementation forgot to add this 3rd parameter.

Lastly, I will add the yfinance package used the same approach as I'm suggesting to resolve the same failure they experienced last week with yahoo source.

bashtage:
#883 builds on this while addressing the issues it created.

bashtage:
Please try master to see if it is working now.

```
pip install git+https://github.com/pydata/pandas-datareader.git
```

If there are positive reports, I'll release very soon.

galashour:
> Please try master to see if it is working now.
> 
> ```
> pip install git+https://github.com/pydata/pandas-datareader.git
> ```
> 
> If there are positive reports, I'll release very soon.

Looks good from my side. Thanks (!)

bashtage:
Fixed in master.  Thanks for the headstart.

PR #890, opened 2021-07-15
Alphabetically order list of supported sources in remote_data.rst file

sugatoray:
Currently the [list of supported sources](https://pandas-datareader.readthedocs.io/en/latest/remote_data.html) in [`remote_data.rst`](https://github.com/pydata/pandas-datareader/blob/main/docs/source/remote_data.rst) is not ordered alphabetically. It will be easier to find out a source if this list is alpahbetically ordered. This PR fixes that.

- Only change made in file: `remote_data.rst`
- This close #890.


bashtage:
Thanks.

PR #929, opened 2022-03-13
Drop row with duplicated index

prakhar-dhakar:
There are cases in yahoo data where there are duplicated rows, this simple fix will remove the second duplicated row if it exists.  A stack overflow thread that contains people facing this error is https://stackoverflow.com/questions/57885790/valueerror-index-contains-duplicate-entries-cannot-reshape-pandas-datareader

- [ ] closes #xxxx
- [ ] tests added / passed


bashtage:
What if the values differ but he index doesn't?  Should PDR return what Yahoo does?

bashtage:
I think you should only drop if the row is identical and the index is identical.

PR #933, opened 2022-05-26
Fixes

### LINKED ISSUE #852: french library that contain operating profit not working
web.DataReader('Portfolios_Formed_on_OP', 'famafrench')

its not working

I tried every data, but only the data that contain OP(operating profitability) is not working.

error message is below
================
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x92 in position 901: invalid start byte

### LINKED ISSUE #932: Key Error (trying to fetch out of range data)
File ~/jupyter_dir/lib/python3.9/site-packages/pandas/core/indexes/base.py:3623, in Index.get_loc(self, key, method, tolerance)
   3621     return self._engine.get_loc(casted_key)
   3622 except KeyError as err:
-> 3623     raise KeyError(key) from err
   3624 except TypeError:
   3625     # If we have a listlike key, _check_indexing_error will raise
   3626     #  InvalidIndexError. Otherwise we fall through and re-raise
   3627     #  the TypeError.
   3628     self._check_indexing_error(key)

KeyError: 'Date'

INTC 1900-01-01 through 1980-03-16, for example.

There is no data for that period. We should get an empty string and catch the error on a try-except routine. Pandas try to read nonexistent data and crashes.

Not a big issue but when trying to release the internet from unnecessary traffic it should count.

In my case, I keep some data for research and only try to fetch whatever data is not in the range already on file.

Best,

Laurentino

stoffprof:
Fixes problems in:
- FamaFrenchReader with encoding in one file
- YahooDailyReader will now return a a DataFrame of nans when a single symbol has no data

- [x] closes #932 
- [x] closes #852
- [ ] tests added / passed

Sorry, I'm kind of a github idiot so I'm sure I'm doing something wrong but these are two pretty quick fixes that don't appear to break anything else. I'm happy to try again if someone can give me guidance about what else needs to be done here.


stoffprof:
@bashtage I'm not sure what these failed tests mean or how to proceed. If you can give any guidance on what needs to be fixed I can take care of it

PR #994, opened 2024-08-08
fix exception for no data with `YahooDailyReader`

Borda:
addresses issue when data are not found in requested page, to reproduce:
```py

# Fetch stock data using pandas_datareader
def fetch_stock_data(symbol='AAPL', start_date='2020-01-01', end_date='2021-01-01'):
    df = web.DataReader(symbol, 'yahoo', start_date, end_date)
    return df['Close'].values

print(fetch_stock_data())
```


pandas_datareader/yahoo/daily.py
```diff
@@ -149,7 +149,7 @@ def _read_one_data(self, url, params):
         try:
             j = json.loads(re.search(ptrn, resp.text, re.DOTALL).group(1))
```
Borda, line 150: if a patter is not found it crashes with different exception:
`AttributeError: 'NoneType' object has no attribute 'group'`

PR #1004, opened 2025-06-05
add dependency of distutils by installing setuptools

ericzhang-debug:
- [✅] closes #xxxx
- [✅] tests added / passed
- [✅] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [✅] passes `black --check pandas_datareader`
- [✅] added entry to docs/source/whatsnew/vLATEST.txt

Add dependency of distutils by installing setuptools, which is not included in standard library in Python 3.12 and above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ Using pandas datareader requires the following packages:
 -   pandas>=1.5.3
 -   lxml
 -   requests>=2.19.0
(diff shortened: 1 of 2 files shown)


ericzhang-debug:
Why it could not pass the check???? I did not modify anything else except requirements.txt.

PR #1011, opened 2026-04-10
fix: restore pandas 3 compatibility

FLIER001:
- [ ] closes #xxxx
- [ ] tests added / passed


bashtage:
pdr has always been pretty unstable. This is why it was originaly removed from pandas. Looking at the test fails, some are due to values changing from providers for historical data. This happens for many reasons, e.g., a historical correction. Happy to have AI assisted patches, but I think some of the fixes need human guidance.

bashtage:
This PR is already getting heavy. I am mostly on board with what SciPy is doing with heavy AI PRs since thye often produce too many wide changes to be practically reviewable.  

To be clear, this isn't a no to this, but it needs to be kept in a range to be reviewed.

PR #1033, opened 2026-08-27
BUG: Preserve SDMX single-dimension label order

### LINKED ISSUE #837: _parse_dimensions() messes up data for non-ordered single indicies
This seems like a bad bug because it messes up your data, and you might not even notice.

If a dimension is not returned in order (e.g. ['1960', ..., '1969', '1950', .... '1959']) and only one index is applicable, your data is mixed. This happens because the `.levels` member of Pandas's MultiIndex contains the indices sorted, no matter which order they were original. For the order, there is a second member: `.codes` (since 0.24.0, not sure how that was handled before).

Basically, this line breaks it: https://github.com/pydata/pandas-datareader/blob/master/pandas_datareader/io/jsdmx.py#L115:
```
if len(arrays) == 1 and isinstance(midx, pd.MultiIndex):
    # Fix for pandas >= 0.21
    midx = midx.levels[0]
```
because it takes the ordered values (e.g., alphanumeric), not the presented order.

To see this in action, try to load the `HISTPOP` data from OECD. If you compare it with this data: https://stats.oecd.org/Index.aspx?DataSetCode=HISTPOP, you can see that pandas-datareader shifts all the data (that is only available from 1960) to 1950, and therefore, it looks like you only have data up to 2010. (If you only request data from the API with a start date after 1959, this does not happen).

Correct:
image

Incorrect (value for 2018 is where value for 2008 should be, etc.):
image

Apparently, this can be fixed by removing the "fix" in line 115 completely (not sure why it is there or if it is still needed) or by changing it to `midx = pd.Index([midx.levels[0][x] for x in midx.codes[0]])` (but that only works for pandas >= 0.24).

aerraj:
## Summary

Fixes #837.

Use `MultiIndex.get_level_values(0)` when flattening a single SDMX dimension. `.levels[0]` sorts the labels independently of the observation values, so unsorted payloads silently attach data to the wrong dates or series. Preserve payload order for both SDMX-JSON v1 and v2.

Add eight cases covering sorted/unsorted dates and series labels, v1/v2 payloads, and missing observations. Correct the existing land-use fixture assertion: its expected values are explicitly keyed to observations 30/31 (1985/1986), not 2010/2011. For example, the raw fixture gives Japan arable land as 53790 in 1985 and 45930 in 2010. The existing expected numbers are retained under their correct dates.

## Verification

- Before the fix: six of the eight new cases failed on upstream.
- `python -m pytest pandas_datareader/tests/io pandas_datareader/tests/test_deprecate_kwarg.py --tb=short`: **21 passed**.
- Black and Flake8 on the changed Python files: passed.
- `git diff --check`: passed.
- Added a release-note entry in `docs/source/whatsnew/vLATEST.txt`.

Tested with Python 3.12, pandas 2.3.3, and NumPy 1.26.4. Tests use local fixtures; live provider/API tests were not run. Single-dimension results now follow payload order; callers needing chronological order should sort the assembled DataFrame (the legacy OECD reader already does this).

- [x] closes #837
- [x] tests added / passed
- [x] changed Python files pass Flake8 and Black


aerraj:
Please review this pull request (˶˃ ᵕ ˂˶)

